PR #3, opened 2023-02-07
Double inclusions and definition

woutersj:
Precompilation isn't working properly for me because of double inclusions (e.g. `CriticalTransitions.jl` includes `StochSystem.jl` and `io/io.jl`, and then `io/io.jl` includes `StochSystem.jl` again).

The functions `central` and `central2` are defined twice.


reykboerner (commented):
Hi Jeroen, could you help me understand these changes please? Indeed the double definition of the finite difference functions `central`, `central2` should be removed and I agree that it's good to put these functions in `utils.jl`. But now it looks like `StochSystem.jl` is not included in many of the files in `dev/` and `src` that use the StochSystem type, since it looks like you commented the respective `include(...)` lines. Is this intended?

woutersj:
Hi Reyk, AFAIK `include(foo.jl)` can be interpreted as simply "insert the source code of `foo.jl` here", so if you issue `using CriticalTransitions`, Julia loads `CriticalTransitions.jl`, which in turn includes `StochSystem.jl` and `io/io.jl`. Since `io/io.jl` again includes `StochSystem.jl`, the module `CriticalTransitions` now contains two copies of the same function. The compiler was complaining about this and it took a long time to load the module for me.
Since the files in `dev` and `src` where I have removed inclusion are part of the same module as `StochSystem`, they will have access to it, since it was already loaded into the module and there should be no need to include it again.

PR #4, opened 2023-02-07
avoid name clash between Symbolics.optimize and Optim.optimize

woutersj:
Both Optim and Symbolics expose a function optimize. This PR qualifies the use of `optimize`.


reykboerner (approved):
Thanks! :)

PR #13, opened 2024-02-19
Add gMAM Meier-Stein notebook

oameye:


oameye:
It would be nice to add the examples/scripts to a documentation too. If you agree, I could attempt this.

PR #19, opened 2024-02-21
Import CoupledODEs such that it gets properly dispatched

oameye:
CoupledODEs from DynamicalSystems was not properly imported such that the method became unavailable in the top namespace. The following errored but now uses the method defined in DynamicalSystems.
```julia
using CriticalTransitions, StaticArrays

function meier_stein(u, p, t) # out-of-place
    x, y = u
    dx = x-x^3 -10*x*y^2
    dy = -(1+x^2)*y
    SA[dx, dy]
end
σ = 0.25

sys = StochSystem(meier_stein, [], zeros(2))

tspan = (0.0, 100.0)
prob = ODEProblem{false}(meier_stein, SA[0.0, 0.0], tspan, ())
kpo = CoupledODEs(prob) # DynamicalSystems method

```


test/stochsystem.jl
reykboerner, line 1: Would be great to turn this into an actual testset that tests the different possible input types of `StochSystem` and `CoupledODEs`.

reykboerner (approved):

reykboerner (approved):

reykboerner:
Thanks for fixing this bug!

PR #20, opened 2024-02-21
Fix `StochSystem(::CoupledODEs)` method

oameye:
The parameter field became `Vector{Vector{...}}` after using the `StochSystem(::CoupledODEs)` method.


reykboerner:
Thanks for spotting and fixing this @oameye!

PR #25, opened 2024-02-23
Refactor gMAM

oameye:


oameye:
@reykboerner What is the reason that we have to re-interpolate the path after every optimization path?

.JuliaFormatter.toml
reykboerner, line 1: Could you please explain what this does?

reykboerner:
Hi @oameye, are there any changes here that change what `geometric_min_action_method` does or are they all just cosmetic in terms of how the code is written? It's a bit difficult to tell from the PR title and by scrolling through the changes.

.JuliaFormatter.toml
oameye: Ha yes sorry. Maybe I shouldn't have committed this. This is a configuration file that format the code to the SciML style guidelines. It works with JuliaFormatter.jl. Feel free to delete this. We could opt to add a CI who automatically formats the code. But I know that Juliadynamics is also still arguing what style they want to use.

oameye:
All the changes should be cosmetic. I just refactored and reformatted the code. The main changes are:
* Refactored the interpolation step into its own function
* Make the two methods dependent on each other, that way we don't have redundant code.

oameye:
All the changes should be cosmetic. I just refactored and reformatted the code. The main changes are:
* Refactored the interpolation step into its own function
* Make the two methods dependend on each other, that way we don't have redundant code.

.JuliaFormatter.toml
reykboerner: Okay, I'll omit it now and we can discuss the CI in our next meeting.

reykboerner:
Great, thanks for the additional description! Merging this now.

Issue #22, opened 2024-02-21
Deal with in-place and out-of-place function in StochSystem


oameye:
To deal with in-place and out-of-place functions in StochSystem we can do something similar as they do in the SciML Ecosystem for SDEProblem or ODEProblem. Something like:
```julia
# Define a parametric type
struct StochSystem{iip}
    f::Function
    pf::Parameters
    u::State
    σ::Float64
    g::Function
    pg::Parameters
    Σ::CovMatrix
    process::Any
end;

function StochSystem{iip}(...) where iip
	...
end
```

reykboerner:
Have you already implemented this in one of your PRs @oameye?

PR #27, opened 2024-02-24
iip or oop functionality

oameye:
add iip or oop kwargs to `simulate` and `relax`


src/StochSystem.jl
```diff
@@ -48,8 +48,8 @@ end
 Concatenates the deterministic and stochastic parameter vectors `pf` and `pg` of a StochSystem `sys`.
 """
 function p(sys::StochSystem)
-    [sys.pf, sys.pg]
-end;
+    isnothing(sys.pg) ? sys.pf : [sys.pf, sys.pg]
```
reykboerner, line 51: I think this would break things the way we write the deterministic part `sys.f`,
```julia
function f(u,p,t)
    a, b, ... = p[1]
    ...
end

pf = [a, b, ...]
sys = StochSystem(f, pf, u0)
```
So `p(sys)` should always return a Vector of Vectors.

The issue with `StochSystem(ds::CoupledODEs)` has been solved with PR #20.

reykboerner (commented):
If it's okay with you I'll undo the changes in `src/StochSystem.jl`, I'm happy to merge the rest.

oameye:
I would merge this once we also changed the Stochsystem struct to allow for iip or oop (issue #22)

oameye:
I would merge this instead into #76

PR #67, opened 2024-03-29
add Downgrade CI

oameye:
https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatibility/110063
https://www.stochasticlifestyle.com/semantic-versioning-semver-is-flawed-and-downgrade-ci-is-required-to-fix-it/


oameye:
Now the compat entries are set very high. We should lower them a bit. Feel free to try some things out and see if the tests fail. @reykboerner @ryandeeley @raphael-roemer

Issue #75, opened 2024-04-21
restructure `StochSystem` to use `SDEIntegrator`


oameye:
As we would like to mimic the `DynamicalSystems` interface, we should transition to use `SDEIntegrator` and workout a similar framework as [continuous_time_ode.jl#L26-L66](https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/v3.7.1/src/core_systems/continuous_time_ode.jl#L26-L66). This will be a big overhaul of the package. This should be done in parallel with #42.

Issue #42, opened 2024-03-27
Dealing with deterministic parameter and stochastic parameter


oameye:
Now the user is expected construct their deterministic function in a way that it takes in parameter of the form `[dp, sp]` with `dp` the deterministic vector of parameters and `sp` the stochastic vector of parameters. However, this makes it dificult to convert deterministic ODE of other packages from DynamicalSystems, OrdinaryDiffEq and HarmonicBalance. Instead, I propose we change handling of the parameters in such a way that the deterministic function `f(u, p, t)` takes in a `Vector` instead of `Vector{Vector}`. 

https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/09bb88dc69fdf2d119e5dfd6e60c064234308abc/docs/src/tutorial.md?plain=1#L27-L35

reykboerner:
Yes, we thought about this as well back then when writing the `StochSystem` struct. The reason why we did this is that the `SDEProblem` from DiffEq only has one parameter field `p` for both the deterministic function `f` and the noise function `g`. However, in the `StochSystem` struct, we wanted to have the deterministic and stochastic parameters separately because the idea is to e.g. take a deterministic system with a set of parameters and throw different noise terms at it which might have different numbers of parameters.

Ideally, `f(u,p,t)` would take in a `Vector` for `p` (and not `Vector{Vector}`) and behind the scenes the `StochSystem` would sort this together with the stochastic parameters `pg` to pass it to `SDEProblem` in a way that still works. Any ideas on how to best do this are very welcome :)

oameye:
> Ideally, `f(u,p,t)` would take in a `Vector` for `p` (and not `Vector{Vector}`) and behind the scenes the `StochSystem` would sort this together with the stochastic parameters `pg` to pass it to `SDEProblem` in a way that still works. Any ideas on how to best do this are very welcome :)

I think we can work this out. I will start a PR

oameye:
I have been looking into how StochasticDiffEq does parameter handling and it seems to opt for one iterator (vector, tuple, etc.) where the functions `f` or `g` know which index to call for a specific param. I would try to mimic this framework such that we have a smooth interplay between ModelingToolkit and StochasticDiffEq.

PR #76, opened 2024-04-21
Stochsystem overhaul

oameye:
Work on #75, #42 and #22


reykboerner:
I've taken a quick look at this and have three comments:

- `CoupledSDEs` now has one parameter field `p` for both functions `f` and `g`, but I still think there should be a constructor function that lets you pass individual parameter fields `pf` and `pg` as input arguments (because otherwise it will be difficult to combine existing `CoupledODEs` drift functions with different noise functions).
- We should also import the `step!` and `set_state!` functions used in DynamicalSystemsBase and then rewrite the callbacks in e.g. `transition` in terms of these functions. This seems to be faster than callbacks.
- It makes sense to me that in DynamicalSystemsBase the function should be called `CoupledSDEs` to match the nomenclature of `CoupledODEs`, but our package focuses more on noise- and rate-induced transitions, so should we still keep an alias like `StochSystem` or `StochasticSystem`?

oameye:
> We should also import the `step!` and `set_state!` functions used in DynamicalSystemsBase and then rewrite the callbacks in e.g. transition in terms of these functions. This seems to be faster than callbacks.

The documentation of DynamicalSystems.jl says that recommend using ODEProblem when using callbacks.
image

oameye:
We should only merge once the CI, Dowgrade and docs are working. Will take a look tonight :)

reykboerner:
The `transition` function throws an error in the tests:
> ERROR: cannot assign a value to imported variable CriticalTransitions.trajectory from module Main

Can check later what the issue is. Also, `trajectory` seems to be a lot slower than `simulate`. Do you know why @oameye ?

src/largedeviations/min_action_method.jl
```diff
@@ -30,7 +30,7 @@ The minimization can be performed in blocks to save intermediate results.
 If `save_info`, returns `Optim.OptimizationResults`. Else, returns only the optimizer (path).
 If `blocks > 1`, a vector of results/optimizers is returned.
 """
-function min_action_method(sys::StochSystem, x_i::State, x_f::State, N::Int, T::Real;
+function min_action_method(sys::CoupledSDEs, x_i::SVector, x_f::SVector, N::Int, T::Real;
```
oameye, line 33: I think it is best to not force to user to use `Svector`

src/CriticalTransitions.jl
```diff
@@ -50,6 +50,7 @@ export equilib, fixedpoints, basins, basinboundary, basboundary
 export simulate, relax
 export transition, transitions
 export fw_integrand, fw_action, om_action, action, geometric_action
+export div_drift
```
oameye, line 53: To keep the namespace minimal, we should limit the amount of functions we export. I don't think we need to export `div_drift`

oameye:
We need to update the docs.

oameye:
dev, examples and ext have still to be updated

src/CriticalTransitions.jl
reykboerner: I think it would be convenient to export because it is quite central to the Onsager-Machlup action. Of course one can compute it pretty easily. We could also consider not exporting `fw_integrand`

oameye: But a user would never call it in a usual workflow right? If they want to call they can just import the function in the namespace or calll it with Criticaltransition.div_drift

docs/src/quickstart.md
```diff
@@ -14,7 +14,7 @@ As this module is not published yet, there are two ways to access it:
 ## Basic usage
 The general workflow of `CriticalTransitions` essentially follows two steps:
 
-1. Define your system (see [Define a CoupledSDE](@ref))
+1. Define your system (see [Define a CoupledSDEs system](@ref))
```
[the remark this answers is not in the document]
reykboerner: Oops, yes.

src/CriticalTransitions.jl
```diff
@@ -16,6 +16,11 @@ using Symbolics
 using Optim, Dierckx
 using Printf, DrWatson, Dates, Statistics
 
+# Define custom types
+Parameters = Union{Vector{Any}, Nothing};
```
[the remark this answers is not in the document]
reykboerner: I agree, I just re-added these custom types because some functions are still dependent on them, which caused an error when they were removed. In the long-run we should follow the style conventions, absolutely.

oameye:
- [ ] check how DynamicalSystems of the state and parameters and it similarly
- [ ] Get docs up to date with this branch
- [ ]  Make extention out of baisins 
- [ ] Make helpful constructors for CoupledSDEs
- [ ] Normalise g function in CoupledSDEs and noise strength field 
- [ ]

oameye:
DynamicalSystems allows the parameters to be of the type that the ODEProblem allows.

oameye:
We will have to add the covariance matrix and noise strength to the CoupledSDEs struct. There is no way to construct these quantities from the SDEIntegrator

oameye:
https://github.com/SciML/DiffEqNoiseProcess.jl/issues/202

oameye:
[ ] add documentation about S in `CoupledSDEs{IIP,..., S}`
[ ] switch p and noise strength in creation of CoupledSDEs

PR #117, opened 2024-10-08
RateSystem

raphael-roemer:
We would like to add functionality for Rate dependent forcing functionality and easy system creation of non-autonomous systems


src/RateSys.jl
```diff
@@ -0,0 +1,53 @@
+#  |                    |                        |                  |
+# t_i    autonomous    t_ni   non-autonomous   t_nf   autonomous   t_f
+#  |                    |                        |                  |
+
+using DynamicalSystems
+using Plots
+
+# we write
+
+function RateSyst(tni,tnf,f,λ,p_λ,initvals)
```
reykboerner, line 10: kugjyfi

src/RateSys.jl
```diff
@@ -0,0 +1,53 @@
+#  |                    |                        |                  |
+# t_i    autonomous    t_ni   non-autonomous   t_nf   autonomous   t_f
+#  |                    |                        |                  |
+
+using DynamicalSystems
```
reykboerner, line 5: It should be sufficient to use `DynamicalSystemsBase`.

reykboerner, line 5: It should be sufficient to use `DynamicalSystemsBase`.

src/RateSys.jl
reykboerner, line 10: In my opinion, these are quite many positional arguments, which makes it less user-friendly. What do you think of reorganising this in the following structure:

```julia
RateSystem(ds::ContinuousTimeDynamicalSystem, rate_protocol::function)
```

where `ds` can be a `CoupledODEs` or a `CoupledSDEs` and `rate_protocol` is your `λ` function that determines how the parameters change over time. This function could have the structure

```julia
rate_protocol(u, p_lambda, t; t_start=0.0, t_end=Inf, kwargs...)
```
i.e. it would contain all the input arguments needed to define the rate forcing function. Behind the scenes, this would then create and return a new `CoupledODEs` or `CoupledSDEs` with the explicit time dependence specified by `rate_protocol`.

reykboerner: The naming here is just a suggestion, but generally I think we should give more descriptive names than just `lambda` or `tnf` (clarity over brevity!)

reykboerner (commented):
Thanks a lot for working on this!

I very much like the idea of starting with the general mathematical formulation `\dot x = f(x(t), lambda(t))`, and have made a suggestion of how this could be reflected more in the way a user would define a `RateSystem`.

One thing I am unsure about is whether we just need a constructor function `RateSystem` that returns either a `CoupledODEs` or `CoupledSDEs`, or whether we need an additional type in order to pass the necessary information about the rate system to the methods we want to add.

Maybe we can brainstorm the methods we would like to add here and think about whether they work with a `CoupledODEs` or whether we need more.

src/RateSys.jl
raphael-roemer: This sounds great, thanks Reyk! However, rate_protocol should not depend on u, in my opinion.

reykboerner:
Hey, what's the status on this @raphael-roemer @ryandeeley ?
If you continue working on this, don't forget to merge the `main` branch into the `RateSystem` branch first, since the main branch has changed in the meantime.

ryandeeley:
> Hey, what's the status on this @raphael-roemer @ryandeeley ? If you continue working on this, don't forget to merge the `main` branch into the `RateSystem` branch first, since the main branch has changed in the meantime.

Hello, Raphael and I met again last week to pick this up again. We agree that it would be nice to pass some used-defined ``auto_sys::CoupledODEs`` and ``\lambda::Function`` - which respectively describe 1) the reference system in an autonomous form, and 2) the time-dependent evolution of the system parameters - to a function ``RateSystem()`` that will return a modified ``nonauto_sys::CoupledSDEs`` describing the associated nonautonomous version of the system. We also agree that it'd be nice if ``RateSystem()`` contains the fewest positional arguments as possible. I've had a crack at writing this generally, and applying to an example case. The user has to provide 
* the base system ``auto_sys::CoupledSDEs``
* the function ``\lambda::Function`` that describes the time-dependent evolution of the system parameters. 
We are proposing that ``\lambda()`` should return **all** system parameters at each time instance, not just those that are explicitly time-dependent, to avoid the (ambiguous) choice of which parameters in the parameter vector of the original system need to be overwritten. Instead, the idea is that the function ``RateSystem()`` - which has ``auto_sys::CoupledODEs`` and ``\lambda::Function`` as positional arguments -  creates a modified time-dependent drift function by evaluating the drift function ``auto_sys.integ.f(u,p,t)`` at ``p=\lambda(t)`` at each time instance. Using the useful function ``remake()``, one then creates a ``ODEProblem`` which is identical to the corresponding ``ODEProblem`` of ``auto_sys::CoupledODEs`` in all components except for the drift function - which becomes the modified time-dependent drift function. The parameter vector and the initial time  of the ``ODEProblem`` also change in this new construction. This ``ODEProblem`` is then converted back to a ``CoupledODEs``, which is ultimately what ``RateSystem()`` then returns. 

As I've written it, the positional arguments of ``RateSystem()`` are 
* ``auto_sys::CoupledODEs``
* ``\lambda::Function``
* ``r::Float64``
* ``t0::Float64``
The first two arguments are hopefully clear. The argument `r` describes a scalar control parameter specifying the scaled rate of change of the parameters - because ultimately we are considering ODEs of the form ``dxₜ/dt = f(xₜ,λ(rt))``. The argument ``t0`` is the initial time that the returned ``CoupledODEs`` takes - which is fairly important in this nonautonomous context. 

There are then keyword arguments ``p_lambda::Vector=Float64[]`` which (in a meta way) contains the parameters for the evolution of the time-dependent system parameters. For instance, the minimal/maximal value of the parameter in a typical ramping scenario. I also find it most logical to have ``t_start::Float64=-Inf`` and ``t_end::Float64=Inf`` as keyword arguments of ``RateSystem()``. These entries allow for a quick modification of the user-provided ``\lambda()::Function`` in order to "clip" the period of non-autonomous parameter evolution - specifically, the modified parameter function will assume the same values ``\lambda(t)`` in the interval ``t\in[t_start,t_end]`` but for ``t < t_start`` or ``t > t_end`` assumes ``\lambda(t_start)`` and ``\lambda(t_end)`` respectively. Note that this clipping of the ``\lambda()::Function`` occurs independently of the rate value ``r`` prescribed: the value of ``r`` has the effect of squeezing/stretching **this particular** clipped parameter evolution. 

Note that there are two non-user-level functions ``modified_drift()`` and ``rate_protocol()`` that ``RateSystem()`` calls in order to construct the modified time-dependent drift function. 

I tested this proposed ``RateSystem()`` on the paradigmatic one-dimensional model for studying R-tipping, and it seems to return the solutions one expects (in reference to the behaviour before / after the critical ``r``-value for R-tipping with the particular time-dependent parameter evolution considered - in this case this is ``r = 4/3``, for R-tipping of the trajectory that in backward-time ``t->-\infty`` limits to the attractor of the "past" autonomous system ``x = -1``). I can also see whether this works nicely for a two-dimensional system. 

I've attached a ``.txt``of the ``.jl`` file where the function ``RateSystem()`` is written up and the tests are made, so let me know what you think of this whenever you get time! 

[RateSystemDraft.txt](https://github.com/user-attachments/files/19275654/RateSystemDraft.txt)

reykboerner:
Hey @ryandeeley, thanks for the update! Could you push your proposed changes to this branch? This is what git is designed for, and pushing here will automatically run the tests. (The idea is that this draft pull request is continuously developed until we are happy to merge it with the main branch)

ryandeeley:
> Hey @ryandeeley, thanks for the update! Could you push your proposed changes to this branch? This is what git is designed for, and pushing here will automatically run the tests. (The idea is that this draft pull request is continuously developed until we are happy to merge it with the main branch)

Yes, we can do that, although I only wrote this yesterday and haven't explicitly discussed it with Raphael yet, who is also drafting some ``RateSystem()`` function. We plan to compare each of our proposals, and I believe we're meeting again tomorrow, so we could push something after then.

test/ratesystem/RateSystemDraft2.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+# convenience functions
+
+RateProtocol(λ::Function,p_lambda::Vector,r::Float64)=RateProtocol(λ,p_lambda,r,-Inf,Inf)
+RateProtocol(λ::Function,r::Float64)=RateProtocol(λ,[],r,-Inf,Inf)
+#RateProtocol(λ::Function,p_lambda::Vector,r::Float64,t_start::Float64)=RateProtocol(λ,p_lambda,r,t_start,Inf)
+#RateProtocol(λ::Function,r::Float64,t_start::Float64)=RateProtocol(λ,[],r,t_start,Inf)	
+
+function modified_drift(u,p,t,ds::ContinuousTimeDynamicalSystem,λ::Function,t_start::Float64,t_end::Float64,r::Float64;
```
reykboerner, line 23: I'm wondering whether this can/should be formulated as an in-place function (i.e. the function updates `ds` without creating a new variable? Then that function could be named something like `apply_forcing!` or so...

reykboerner:
- [ ] add docstring for `RateSystem` and `RateProtocol`

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,80 @@
...
+end;
+
+"""
+    RateSystem(sys::ContinuousTimeDynamicalSystem, rp::RateProtocol, t0=0.0; kwargs...)
+
+Applies a time-dependent [`RateProtocol`](@def) to a given autonomous dynamical system
+`sys`, turning it into a non-autonomous dynamical system. Returns a [`CoupledODEs`](@ref)
+with the explicit parameter time-dependence incorporated.
+"""
+function RateSystem(
```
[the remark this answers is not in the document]
reykboerner: Yes, sounds good to me!
We should only have the `!` in the function name if it really updates the existing CoupledODEs/SDEs. If it creates a new version of it (which I think it does?) we should call it `apply_ramping`.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,80 @@
...
+    auto_sys::ContinuousTimeDynamicalSystem, rp::RateProtocol, t0=0.0; kwargs...
+)
+    # we wish to return a continuous time dynamical system with modified drift field
+
+    f(u, p, t) = modified_drift(
+        u, p, t, auto_sys, rp.λ, rp.t_start, rp.t_end, rp.r; kwargs...
+    )
+    prob = remake(auto_sys.integ.sol.prob; f, p=rp.p_lambda, tspan=(t0, Inf))
+    nonauto_sys = CoupledODEs(prob, auto_sys.diffeq)
```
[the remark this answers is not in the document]
raphael-roemer: good point. For now, I fixed it by only allowing `CoupledODEs` as inputs, but I will have a look at whether and how this works for `CoupledSDEs`

raphael-roemer:
@Datseris, what do you think about the setup of the r-tipping functionality? It would be great to get your feedback on this!

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+
+If `t_start` and `t_end` are not provided, they are set to `t_start=-Inf`, and `t_end=Inf`.
+If `p_lambda` is not provided, it is set to an empty array `[]`.
+"""
+mutable struct RateProtocol
+    λ::Function
+    p_lambda::Vector
+    r::Float64
+    t_start::Float64
+    t_end::Float64
```
oameye, line 37: This struct should be made concrete. Both `p_lambda` and `λ` are non concrete types instead one should use Parametric types. Also, `RateProtocol` should probably not be `mutable` as this would lead to bad performance. Instead I propose
```julia
struct RateProtocol{F, T}
    forcing_function::F
    p_lambda::Vector{T}
    rate::T
    time_interval::Tuple{T,T}
end
```

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+
+The RateProtocol contains all the fields (information) that allows to take an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ)` 
+with `λ ∈ Rᵐ` containing all system parameters, and make it an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ(rt))`
+with `λ(t) ∈ Rᵐ` constant before time `t_start` and also after time `t_end`. 
+
+RateProtocol contains the following fields:
+- `λ::Function`: forcing function of the form `λ(p, t_start; kwargs...)``
+- `p_lambda::Vector`: parameters of forcing function
```
oameye, line 21: p_lambda::Vector`: parameters of the forcing function

Datseris:
hi, i was on leave, having some personal life problems, i'm slowly getting back and i'll have a look at this over the next 4 days.

Datseris:
> @Datseris, what do you think about the setup of the r-tipping functionality? It would be great to get your feedback on this!

@raphael-roemer can you please clarify: this PR does not have any functionality for r-tipping, it only defines a construct (RateSystem) that will help develop r-tipping functionality in the future, correct?

docs/src/examples/RateSystem.md
```diff
@@ -0,0 +1,97 @@
+# Studying R-Tipping
+
+Let us explore a simple prototypical example of how to use the R-tipping functionality of this package.
```
Datseris, line 3: R-tipping isn't defined anywhere until here. Probably best to call this "rate dependent tipping".

On a side-note: there is no educational resources (to my knowledge) that properly teaches R-tipping to a non-PhD. The best attempt is a paper from Ritchie et al 2023, but it is still advanced (research article). I am currently working on one which will come in v2 of my textbook. I will try to upload some notes on GitHub in the meantime to cite here.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+# We ask the user to define: 
+#  1) a ContinuousTimeDynamicalSystem that should be investigated and
+#  2) a protocol for the time-dependent forcing with the struct RateProtocol
+
+# Then we give back the ContinuousTimeDynamicalSystem with the parameter 
+# changing according to the rate protocol
+"""
+    RateProtocol
+
+The RateProtocol contains all the fields (information) that allows to take an ODE of the form
```
Datseris, line 13: If you want to be harmonious with the rest of DynamicalSystems.jl and DIfferentialEquations.jl, you need to replace λ with p throughout everything. The parameter container is always called `p` in both ecosystems.

I strongly recommend you do that because the choice of how to name this `λ` is completely arbitrary. So if there is any reason to prefer one option, you should do it. Here especially it is a very good reason.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+# here λ = λ(t) ∈ Rᵐ is a function containing all the system parameters 
+
+# We ask the user to define: 
+#  1) a ContinuousTimeDynamicalSystem that should be investigated and
+#  2) a protocol for the time-dependent forcing with the struct RateProtocol
+
+# Then we give back the ContinuousTimeDynamicalSystem with the parameter 
+# changing according to the rate protocol
+"""
+    RateProtocol
```
Datseris, line 11: I would recommend to rename this to `RateConfig`. Config, from configuration, is the name give typically in Julia to such composite types whose purpose is to configure some functionality. I like a lot that this is a dedicated struct, instead of keyword arguments to a function. But I believe naming it Config also makes it more harmonious with DynamicalSYstems.jl which follows this convention in many packages.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+
+The RateProtocol contains all the fields (information) that allows to take an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ)` 
+with `λ ∈ Rᵐ` containing all system parameters, and make it an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ(rt))`
+with `λ(t) ∈ Rᵐ` constant before time `t_start` and also after time `t_end`. 
+
+RateProtocol contains the following fields:
```
Datseris, line 19: The user shouldn't care about what fields the data structure has, which also allows you to attribute more private fields to it. Here you should just make a section called `## Keyword arguments` in harmony with the rest of the ecosystem.

Also, refer to the data structure in code since it is a code name (type). Use backticks and write `RateProtocol` instead of RateProtocol.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+- `λ::Function`: forcing function of the form `λ(p, t_start; kwargs...)``
```
Datseris, line 20: We have a couple of issues here. First, it is stated wrongly. It is `λ(p, t; kw...)` it is not a function of the starting time, it is a function of _time_ in general. You can see this clearly in the source code. Second, what is the purpose of `kwargs`? The user makes the function themselves. They don't need `kwargs`. If they want to capture variables inside this function they can make a closure. I think `kwargs` unecessarily complicates this, especially because it is not discussed at all in the docstring here.

Also, is `p` supposed to be `p_lambda` from below? If yes, why not give the same name? (1 word per concept rule)

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+- `p_lambda::Vector`: parameters of forcing function
+- `r::Float64`: rate parameter
+- `t_start::Float64`: start time of protocol
+- `t_end::Float64`: end time of protocol
+
+Default values 
+==============
+
+If `t_start` and `t_end` are not provided, they are set to `t_start=-Inf`, and `t_end=Inf`.
+If `p_lambda` is not provided, it is set to an empty array `[]`.
```
Datseris, line 30: You can avoid all of this by simply assigning values to when you stated the keywords. E.g., `t_end = Inf`.

src/r_tipping/RateSystem.jl
Datseris: Also, no need to declare the fields of the arguments.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+The RateProtocol contains all the fields (information) that allows to take an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ)` 
+with `λ ∈ Rᵐ` containing all system parameters, and make it an ODE of the form 
+    `dxₜ/dt = f(xₜ, λ(rt))`
+with `λ(t) ∈ Rᵐ` constant before time `t_start` and also after time `t_end`. 
+
+RateProtocol contains the following fields:
+- `λ::Function`: forcing function of the form `λ(p, t_start; kwargs...)``
+- `p_lambda::Vector`: parameters of forcing function
+- `r::Float64`: rate parameter
```
Datseris, line 22: What is this? Why is it different from `p_lambda`?

src/r_tipping/RateSystem.jl
Datseris: I also prefer the `time_interval` field instead of tstart/end.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+Computing trajectories of the returned [`CoupledODEs`](@ref) can then be done in the same way as for any other [`CoupledODEs`](@ref).
+"""
+function apply_ramping(auto_sys::CoupledODEs, rp::RateProtocol, t0=0.0; kwargs...)
+    # we wish to return a continuous time dynamical system with modified drift field
+
+    f(u, p, t) = modified_drift(
+        u, p, t, auto_sys, rp.λ, rp.t_start, rp.t_end, rp.r; kwargs...
+    )
+    prob = remake(auto_sys.integ.sol.prob; f, p=rp.p_lambda, tspan=(t0, Inf))
+    nonauto_sys = CoupledODEs(prob, auto_sys.diffeq)
```
Datseris, line 96: It is bad practice to access internal fields of structs, especially when there are tested and exported API functions for it. Instead of doing ds.integ.sol.prob, do `referrenced_sciml_prob(ds)`. See: https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/main/src/core/dynamicalsystem_interface.jl#L136

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    p̃ = if r*t ≤ t_start
+        λ(p, t_start; kwargs...)
+    elseif t_start < r*t < t_end
+        λ(p, r*t; kwargs...) # the value(s) of λ(rt)
+    else
+        λ(p, t_end; kwargs...) # the value(s) of λ(rt)
+    end; # the value(s) of λ(rt)
+    return ds.integ.f(u, p̃, t)
```
Datseris, line 71: It is bad practice to access internal fields. Use instead `dynamic_rule(ds)` to obtain `ds.integ.f`.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,98 @@
...
+    r::Float64;
+    kwargs...,
+)
+    if t_start > t_end
+        error("Please ensure that t_start ≤ t_end.")
+    end
+
+    p̃ = if r*t ≤ t_start
+        λ(p, t_start; kwargs...)
+    elseif t_start < r*t < t_end
```
Datseris, line 66: This is confusing to me. If the "rate syste" is "on" between `t_start` and `t_end`, I would expect this to be true. BUt in the source code it isn't. It is not `t` that is compared to `t_start/end`, but rather `r*t`. I think the docs have to start discussing the motivation behind the setup a bit more transparently.

Datseris (changes requested):
@raphael-roemer I am giving a surface/superficial review for now, because more information is necessary for a proper interface review. Can you please state the rational behind the approach here? Why we have both `r` and also dedicated parameters for `lambda`? Why is it that `r*t` is compared with the time instead of just `t`? Just write a couple of sentences of how you envision people (e.g., yourself) using this type. What is the purpose for both `r` and these extra `p_λ` parameters, and what is "time" for the rate system. Is it still `t`, as I would expect, or is it `r*t`, as it is in the source code?

I think `r` is causing problems because it fundamentally changes the nature of time, while it should have been just another parameter, such as an element of `p_λ`.

reykboerner:
Thanks a lot for your comments @Datseris! I think you raise important scientific design questions.

> 1. There are two key parameters to consider, `Δp, Δt`

I agree that two parameters give more info than one, but for nonlinear ramping even that doesn't tell you all about the rate, so the full forcing function is needed anyway. I see the `r` parameter more as a scaling parameter (like the noise strength in large deviation theory) that helps defining criticality and forcing regimes (fast/slow). One could argue that what matters it the time-integrated overshoot of a bifurcation point (which would be `Δp*Δt` in the linear case).

> 2\. Both are behaviors that often are talked about in "r-tipping" literature

True, both are different scenarios - here we discussed using the definition of R-tipping as "failure to track (a time-dependent state in state space, typically a moving equilibrium)". I like the generality of this definition and it includes both behaviors you describe. I also see it as a benefit that the definition does not talk about bifurcations (which are very non-trivial in non-autonomous systems).

> Furthermore we can add more constrains and allow only for specific functions ? linear, tanh? (also allow user-provided function `ratefunction(Δt, Δp, t_start, t_end)` as a possibility, but most people would either use linear or tanh I feel)

Isn't it best to define the core struct very generally and then define convenience contructors for linear/tanh ramping that users would commonly use?

These are my thoughts - @raphael-roemer @ryandeeley @oameye please feel free to add yours :)

And hope things are getting better soon @Datseris!

raphael-roemer:
> > @Datseris, what do you think about the setup of the r-tipping functionality? It would be great to get your feedback on this!
> 
> @raphael-roemer can you please clarify: this PR does not have any functionality for r-tipping, it only defines a construct (RateSystem) that will help develop r-tipping functionality in the future, correct?

Hi @Datseris, thank you very much for your comments! 
You are right, it provides a construct that will help develop r-tipping functionality in the future.

Datseris:
Actually the function `is_basin_transferable(A, mapper; kw...)` should be in this package. I will contribute it via a PR once I have some spare time.

ryandeeley:
> I have some decently organized notes that I will be finishing soon and I can share with you. There the way I am teaching rate tipping is as follows:
> 
>     1. There are two key parameters to consider, `Δp, Δt`, which is the overall change in a parameter and the time window over which it happens. There is a specific reason to take both into account: the first decides whether you cross a bifurcation point or not, while the second decides the rate.

Can `\Delta t = \infty`? Otherwise you are necessarily restricting to systems whose time-dependent dynamics occurs over a finit time-interval. If however you allow for `\Delta t = \infty`, which would correspond to the non-truncated hyperbolic tangent ramping, that together `\Delta t` and `\Delta p` doesn't say anything about the rate of change. For that, you specifically need `r`.   

> 
>     2. Whether the change is reversed or not. If the change is not reversed, and `Δp` does not cross a bifurcation point, we talk about rate tipping. If the change is reversed, and `Δp` does cross a bifuircation point, then we talk about overshooting return. Both are behaviors that often are talked about in "r-tipping" literature, despite being fundamentally different phenomena.
> 
> 
> I am wondering whether it makes sense to take these into account into this PR? Have one more option for the reversal of the time variability?

I'm not familiar enough with R-tipping overshoot studies to comment about useful there, but I agree that it'd be nice if such overshoot scenarios could be simulated using this functionality (then it gets a bit intricate constructing a candidate `λ`-function).

Datseris:
> Can `\Delta t = \infty`? Otherwise you are necessarily restricting to systems whose time-dependent dynamics occurs over a finit time-interval. If however you allow for `\Delta t = \infty`, which would correspond to the non-truncated hyperbolic tangent ramping, then together `\Delta t` and `\Delta p` don't say anything about the rate of change. For that, you specifically need `r`.

I would disagree here. Technically a tanh occurs always over an infinite time interval. In reality however this is not the case and one can define very well an interval over which the ramping occurs. Here is an example: 
```julia
function sigmoid_expression(T, left, right, scale, reference)
    return left + (right - left)*(1 + tanh(2(T - reference)/(scale)))*0.5
end
```

In this expression `T` is time and the function returns the value of hte parameter given time. `left` and `right` are the values of the tanh before and after the ramping. So for us `left = p0` and `right = p0 +dp`. `reference` and `scale` are also both directly related with `t0, dt` with `reference = midpoint = t0 + dt/2` and `scale = dt` directly. If you plot the function I have listed above you will see that no matter what values you choose for left right scale and reference, you will always get a consistent result where the ramping is effectively only acting within a `dt` window. Sure, it may act a little before and after, but we are talking here about a generic framework that is applicable at least approximately to the majority of cases. The effective ramping time `dt` applies here despite technically the ramping window being infinite.

ryandeeley:
> > Can `\Delta t = \infty`? Otherwise you are necessarily restricting to systems whose time-dependent dynamics occurs over a finit time-interval. If however you allow for `\Delta t = \infty`, which would correspond to the non-truncated hyperbolic tangent ramping, then together `\Delta t` and `\Delta p` don't say anything about the rate of change. For that, you specifically need `r`.
> 
> I would disagree here. Technically a tanh occurs always over an infinite time interval. In reality however this is not the case and one can define very well an interval over which the ramping occurs. Here is an example:
> 
> ```julia
> function sigmoid_expression(T, left, right, scale, reference)
>     return left + (right - left)*(1 + tanh(2(T - reference)/(scale)))*0.5
> end
> ```
> 
> In this expression `T` is time and the function returns the value of hte parameter given time. `left` and `right` are the values of the tanh before and after the ramping. So for us `left = p0` and `right = p0 +dp`. `reference` and `scale` are also both directly related with `t0, dt` with `reference = midpoint = t0 + dt/2` and `scale = dt` directly. If you plot the function I have listed above you will see that no matter what values you choose for left right scale and reference, you will always get a consistent result where the ramping is effectively only acting within a `dt` window. Sure, it may act a little before and after, but we are talking here about a generic framework that is applicable at least approximately to the majority of cases. The effective ramping time `dt` applies here despite technically the ramping window being infinite.

src/r_tipping/RateSystem.jl
Datseris: 1) there is no parameter change outside `Δt`, that's the reasoning behing having a `dt`. This doesn't forbid infinite `dt`, as long as `dp` is also infinite. This would the scenario of a perpetually drifting system where the parameter always increases forever. 

2) This is the same fallacy as the argument of @reykboerner . We are comparing whether _only_ `r` is the appropriate generalization, or whether `Δp + Δt` is a better generalization for a generic framework for generic dynamical systems and generic rate tipping, that can cover all 6 scenarios I outline in the book excerpt above, all of which are mentioned as "rate tipping" in the literature. We are not comparing whether specifying the whole function `λ` is better than specifying `Δt + Δp`, that's simply not the comparison here. You always have to specify the full function `λ`, or at least choose from a set of predefined options like linear or tanh anyways.

Yes, you are correct, everything is equivalent - it has to be. My argument is: how fast you change the parameter is only one component of the story, the other component is whether you cross any bifurcation points or not. At least, this story matters for all these overshooting scenarios that people keep classifying under "rate tipping" whether they are the same thing or not. If you care about these scenarios then having `dp` as an input is better, that is my argument. In my setup `Δt` is equivalent with your rate, they are the inverse of each other. You just don't have to do the computation of passing `r` through `λ` to get the parameter change, `Δp = λ(r*dp)`, because here `Δp` is one of the specified inputs. Naturally, everything is equivalent, we have 7 quantities, `t_start, t_end, dt, p_start, p_end, dp, r`. Given a form of `λ`, one only has to specify 4 of these quantities, then the other 3 are derived, no matter which 4 are specified originally. This is guaranteed. The question is: which 4 quantities the better ones to base the struct `RateConfig` on?

So, I am trying to shift the focus away from the `λ` and into the `dp`. And the reason is because I believe at least that this is more generic. Indeed, given a dynamical system a `dp` will tell you whether you will cross a bifurcation point or not independently of what `λ` is chosen as. The assxumption here is that `λ` is always some sort of ramping function that ramps within some interval `dt` and ramps from `p0` to `p0 + dp`.

p.s.: nothing excludes the possibility of allowing several different convenience constructors for the RateCOnfig. You could have one that is based on `r`, and another that is based on `dt + dp`. I am arguing on what the canonical internal representation should be. As a user, you don't care at all about this conversation!

raphael-roemer:
Hello, sorry for the late response, I was on holiday and needed some time to think about how to combine all the good suggestions. 
First, thank you very much for your points! 
I agree that it would be great also to have the possibility to define some ramping via a time interval `dt` and a parameter shift amount `dp` in certain cases. However, I think that this should be done via convenience functions that allow doing this for specific cases (linear ramping, tanh-ramping), and these convenience functions then translate this setup to the current (more generic) setting, which runs behind the scenes. 
I think this would be best as it would allow experts to work in the current, more generic setting, which is as close to the theory as possible, while allowing easy implementation of the usual settings.

For complicated parameter rampings, I don't think that a definition of the `RateConfig` only via `dp` and `dt` is easy to handle:
1) It would require knowing the maximum and minimum of the parameter ramping. This would be attainable, but would require an extra step, and they may depend non-trivially on the parameters of the function that describes the parameter ramping.
2) It would not even necessarily be the same `dp` and `dt` that users would like to use in overshoot settings: a parameter ramping might start far away from the bifurcation threshold and only overshoot the threshold for some short time, and a small amount, i.e., the `dt_overshoot` and `dp_overshoot ` describing the threshold exceedence time and amount could be smaller than `dt` and `dp` describing the full parameter ramping.

raphael-roemer:
Hi, I have now tried to implement what we discussed in the meeting on Friday. Please let me know your thoughts.

I imposed that any parameter ramping function that the user defines needs to fulfill `p(t=-10) =0` and `p(t=10)=1`. Then we always use the function `p(t)` on this time interval ``(t=-10, t=10)`` and scale its amplitude using `dp`, or stretch/compress its duration via `ramp_t_length`.

Datseris:
what is special about 10 and -10?

raphael-roemer:
Nothing is special about these values. I just thought that one has to specify *some* part of the function `p(t)` that should be considered for the ramping. For example, only specifying `p(t)=tanh(t)` would not make clear which part of `p(t)` we take (e.g., tanh((-100,-90)) would be boring). 

Once we fix a time interval like `(-10, 10)`, then one can stretch/compress the function `p((-10, 10))` via changing `ramp_t_length`.  For example, choosing `ramp_t_length=100` and `t_start=-50` would spread out the ramping specified by the function `p((-10, 10))` over the time interval `(-50, 50)`.

Maybe there is a more convenient way to achieve this (e.g., we could ask the user to specify on which interval they want the function `p(t)` to be considered for the ramping), so please let me know your ideas.

raphael-roemer:
Thank you for checking this so quickly! You are right, asking the user to do the stretching/squeezing would also be an option. However, when coding this yesterday, I found the stretching/squeezing the most difficult part to code. So I thought if the user only has to fiddle around with Desmos or pen and paper once to find the right function parameters fulfilling `p(t=-10) = 0 and p(t=10) = 1`, this might be easier than hard-coding the stretching/squeezing. But maybe I am missing something here...

ryandeeley:
> Thank you for checking this so quickly! You are right, asking the user to do the stretching/squeezing would also be an option. However, when coding this yesterday, I found the stretching/squeezing the most difficult part to code. So I thought if the user only has to fiddle around with Desmos or pen and paper once to find the right function parameters fulfilling `p(t=-10) = 0 and p(t=10) = 1`, this might be easier than hard-coding the stretching/squeezing. But maybe I am missing something here...

I think if someone has the function defined across t \in (-10,10) there is a function-independent way of generalising this into t \in (a,b), which could be applied to (t_start,t_start+t_ramp_length) for this case. So maybe it's not really much extra work at all? Especially if we gave a clear example with the tanh function...

ryandeeley:
> What about this?: The user gives a `p(t)` and some time interval `(t_pstart, t_pend)` on which `p(t)` is _monotonically increasing_. We subtract `p(t_pstart)` from `p(t)` and then multiply it by `1/p(t_pend)` to achieve that `p(t = t_pstart) = 0 and p(t = t_pend) = 1` are fulfilled. Then we stretch/squeeze `p(t_pstart, t_pend)` into `(t_start, t_start+t_ramp_length)`. The user can then vary the speed of the ramping via `t_ramp_length`, and the amount of forcing via `dp`.
> 
> In this way, the user would not have to do the stretching/squeezing, but just has to specify a start and end point `(t_pstart, t_pend)` of the interval of the function `p(t)` they want to consider for the ramping.

I think the ingredients you've suggested for what the user has to specify is the best here (it's not actually too far from we already had), but the details still have to be worked out...Let me get back to you later (I'm only on my phone right now and don't have any pen/paper to do some quick check with the maths...)

ryandeeley:
> > I would then propose we incorporate the "change in amplitude" element into this by considering `p̃(t) = p0 + dp*(p(t)-p(t_pstart)) / ( p(t_pend)-p(t_pstart) )`, such that the function _always_ starts at `p̃(t_pstart) = p0` and `p̃(t_pstart)-p̃(t_pstart) = dp`.
> 
> I think that's perfect! Thanks :)
> 
> > I'm not sure we need to restrict to monotonically increasing functions here. I think would just have to change the denominator to `abs(p(t_pend)-p(p_pstart))`. That way if `p(t_pstart) > p(t_pend)` the function `p̃(t) = (p(t)-p(t_pstart)) / abs(p(t_pend)-p(t_pstart))` just satisfies `p(t_pstart) = 0` and `p(t_pstart) = -1`.
> 
> Good point, decreasing is also fine. And regarding monotonicity: If this isn't the case, then it can happen that `p(t_pend)` is not the most extreme forcing value. But we leave this choice to the user, right? Then we just tell the user that if we have monotonicity, `dp`gives the forcing amplitude and without monotonicity, it is just a scaling of the forcing amplitude.

Yeah, I agree with this, because monotonicity is required for the interpretation of `dp` to be the maximal difference in parameter values, otherwise it's just a scaling as you say.

ryandeeley:
Something we also have to resolve is that this requires `p(t_pend)-p(t_pstart)=0`. For the parameters in the system which are not time-dependent, what do we do?

ryandeeley:
So I've just pushed some changes to the `RateSystem.jl` file, which answer to most of the things we discussed yesterday. In particular I've added `t_pstart` and `t_pend` fields to the `RateConfig`, and the `dp`-field has the effect as spelled out yesterday. The `t_ramp_length` has remained a field of `RateConfig`, although there is a correspondence between `t_ramp_length` and what was originally `r` via `r = (t_pend-t_pstart)/t_ramp_length` or equivalently `t_ramp_length = (t_pend-t_pstart)/r` [plus the necessary shiftings of time, because the nonautonomous dynamics is now programmed to always take place within `t \in (t_start, t_start + t_ramp_length)`. If one chooses `t_start = -t_ramp_length/2` and `t_pend = -t_pstart`, this shifting of time is natural.]. 

I haven't yet altered any test files in correspondence with this yet, so that's still to do...

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,122 @@
...
+Time-dependent forcing protocol containing the information to apply a parameter shift to an autonomous system.
+
+Fields
+==============
+
+- p: monotonic function which describes the time-dependent parametric forcing 
+- p_parameters: the vector of parameters which are associated with p
+- t_pstart: the parameter values of the past limit system are given by p(p_parameteters,t_pstart)
+- t_pend: the parameter values of the future limit system are given by p(p_parameters,t_pend) 
+- t_start: the explicit value of time at which the nonautonomous dynamics starts
```
reykboerner, line 22: Hi, it seems redundant (and a bit confusing) to have all these different time inputs (`t_pstart`, `t_start`, ...). I am worried that we are making things too complicated.

To me, the most straight forward would be: you define a forcing function `p(t)` and a tuple `(p_start, p_end)`. For all `t`, the parameters are then given by:
- `p(t_start)` for $t \leq t_{start}$
- `p(t)` for $t_{start} < t < t_{end}$
- `p(t_end)` for $t \geq t_{end}$

This ensures the forcing is continuous. Its derivative might not be, but okay. If you have a tanh forcing on `[t_start, t_end]`and want the past and future limits to be closer to the 'true' limits of the tanh, just widen the interval `[t_start, t_end]`.

reykboerner:
Regarding the normalization of the forcing function: what is your solution for the case that $p(t_{start}) = p(t_{end})$ as @ryandeeley pointed out?

My opinion is: it's not necessary to normalize. What is now called `dp` can still be a scaling, and if you define your function `p(t)` such that it varies from `0` to `1`, then `dp` corresponds to the ramping magnitude.

src/r_tipping/RateSystem.jl
reykboerner: P.S. I am thinking that one applies the scaling (`dp` and `rate`) to the forcing first and then evaluates the start and end values of that rescaled forcing function using the specified `t_start` and `t_end`.

Datseris:
why are you not making `dp` just be an input to the user provided function so that the user explicitly takes care of what `dp` means?

src/r_tipping/RateSystem.jl
ryandeeley: I would disagree that it's redundant and will try to dispel any possible confusion. One needs `p`, `p_parameters`, `p_pstart` and `t_pend` to define a piecewise constant function that the R-tipping experiments are taken with respect to. You need to fix a `t_pstart` and `t_pend` to fix the past and future limit systems. These have to be finite, because `[t_pstart,t_pend]` is rescaled into the finite interval `[t_start,t_start+t_ramp_length]`. If you simply evaluate a rescaled forcing function at some `p_start` and `p_end` each time, then this is going to change the past and future limit systems depending on the rescaling (and could do drastically if you consider a very shallow `rate`).   

The parameter `t_start` is in my opinion unnecessary, since we could always fix `t_start = -t_ramp_length/2` and have that the nonautonomous dynamics occurs in a symmetric window around `t = 0`.

ryandeeley:
> Regarding the normalization of the forcing function: what is your solution for the case that p ( t s t a r t ) = p ( t e n d ) as @ryandeeley pointed out?

In such cases no scaling is performed to those parameters - see the `p_piecewise_scaled` function. Since we have restricted to monotonic functions, here this only applies to fixed (time-independent) parameters. 

> 
> My opinion is: it's not necessary to normalize. What is now called `dp` can still be a scaling, and if you define your function `p(t)` such that it varies from `0` to `1`, then `dp` corresponds to the ramping magnitude. 

I agree that the normalisation is specific here, and to reiterate it assumes monotonicity. The problem with simply scaling the whole function, i.e. `c p(t,...)` is that this will also scale the value associated with the past limit system unless `p(0)` starts at `0`, but why would we want to restrict to such cases? This would require one to incorporate the true starting value of the parameter somewhere else into the code, which might be an option but it doesn't sound so straight forward or at least would require more thinking... The problem also with this scaling is that it would change the values of the fixed time-independent parameters (which is again problematic and seemingly not the intention). 

The way it works right now is admittedly a specific case but it captures the idea that `dp` scales the difference in parameter values of the time-dependent parameters over the interval of time-dependent dynamics. 

Perhaps there is no simple universal and helpful interpretation of "scaling" the amplitude of the parameter functions, which I think has been part of the issue in unanimously understanding this so far... Maybe George's idea to leave the interpretation of the `dp` scaling to the user is the best compromise...

Datseris:
I think we need one more video call; for me at least it is difficult to track the discussion here across different line-comments and different source code files.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,122 @@
...
+- dp: the difference in parameter values attained across the ramping
+
+Default values 
+==============
+
+- t_pstart = -100
+- t_pend = 100
+- p_parameters = []
+- t_start = -t_ramp_length/2
+- dp = 1
```
Datseris, line 33: I don't think these defaults make sense. Unless there is a singular unambiguous value, like 0 for starting time, you should not have defaults and request the user to provide them. 
`p_parameters` should be `nothing` by default as with the rest of DynamicalSYstems.jl

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,122 @@
...
+    kwargs...,
+)
+
+    q(t) = p_piecewise_scaled(p,p_parameters,t_pstart,t_pend,dp,t)
+
+    time_shift = ((t_pend-t_pstart)/t_ramp_length)*(t-t_start)+t_pstart # such that [t_start,t_start+t_ramp_length] is shifted into [t_pstart,t_pend]
+
+    q̃ = q(time_shift)
+
+    return dynamic_rule(ds)(u, q̃, t)
```
Datseris, line 92: what do you do for in-place systems here?

Datseris:
i'm unsubscribing so that i don't get constant updates here, tag me with `@datseris` when I should review again!

docs/src/examples/RateSystem.md
```diff
@@ -0,0 +1,101 @@
...
+Now, we want to explore a non-autonomous version of the system by applying a parameter ramping. 
+As discussed, we consider a setting where in the past and in the future the system is autnonomous and in between there is a non-autonomous period `[t_start, t_start+ramp_t_length]` with a time-dependent parameter ramping given by the function ``p(t)``. Choosing different values of the parameter `ramp_t_length` allows us to vary the speed of the parameter ramping.
+
+We start by defining the function `p(p_parameters, t)`:
+```@example RateSystem
+function p(p_parameters, t)
+    p_max = p_parameters[1]
+    p_ = (p_max/2)*(tanh(p_max*t/2)+1)
+    return SVector{1}(p_)
+end
```
Datseris, line 40: I don't agree with this design. I think the function `p` should not have parameters. It should just be `p(t)`. Also, we should name it something else, as it missleadingly represents the universal symbol for parameters, even though it isn't a parameter at all. It is just a profile. We should name it `profile(t)`.

Datseris: What's the reason for the parameters? It defnitely isn't to control the rate/steepness of the profile, because this is controlled externally with the dp, dt arguments.

Datseris:
I am sorry I have dropped the ball here, right? Should I have been reviewing this but forgot to do so?

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+"""
+    RateSystem(sys::ContinuousTimeDynamicalSystem, rc::RateConfig, pidx; kwargs...)
+
+Creates a `RateSystem` type from an autonomous dynamical system `sys` and a time-dependent
+parametric forcing protocol of `RateConfig` type.
+
+## Keyword arguments
+- `forcing_start = RateConfig.section_start`: Time when parameter shift starts (before this, the resulting system will be autonomous)
+- `forcing_length = RateConfig.section_end - RateConfig.section_start`: Time-interval over which RateConfig.pfunc([RateConfig.section_start, RateConfig.section_end]) is spread out (for window_length > RateConfig.section_end - RateConfig.section_start) or squeezed into (for window_length < RateConfig.section_end - RateConfig.section_start)
+- `forcing_scale = 1.0`: Amplitude of the ramping. The ramping is then automatically rescaled
```
Datseris, line 45: Is this `Δp`?

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+
+"""
+    RateSystem(sys::ContinuousTimeDynamicalSystem, rc::RateConfig, pidx; kwargs...)
+
+Creates a `RateSystem` type from an autonomous dynamical system `sys` and a time-dependent
+parametric forcing protocol of `RateConfig` type.
+
+## Keyword arguments
+- `forcing_start = RateConfig.section_start`: Time when parameter shift starts (before this, the resulting system will be autonomous)
+- `forcing_length = RateConfig.section_end - RateConfig.section_start`: Time-interval over which RateConfig.pfunc([RateConfig.section_start, RateConfig.section_end]) is spread out (for window_length > RateConfig.section_end - RateConfig.section_start) or squeezed into (for window_length < RateConfig.section_end - RateConfig.section_start)
```
Datseris, line 44: Is this `Δt`...? I am confused. Why is this required as a keyword? Given to what it is equal, it isn't even `Δt`. Yet there is no keyword asking for `Δt`. This is so confusing.... why ask again for an information already provided

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+- `forcing_scale = 1.0`: Amplitude of the ramping. The ramping is then automatically rescaled 
+- `t0 = 0.0`: Initial time of the resulting non-autonomous system (relevant to later compute trajectories)
+
+## Description
+The returned `RateSystem.system` is is
```
Datseris, line 49: You shouldn't be referring to `RateSystem.system` here. It isn't valid syntax either, as `RateSyste` is a type and does not have a field `system`. Regardless, just say `rs = RateSystem(...)` and refer to `rs` exclusively. It is a dynamical system because we extended the proper API.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+        return p0
+    else
+        if t < forcing_start+forcing_length
+            # performing the time shift corresponding to stretching/squeezing 
+            time_shift =
+                ((section_end-section_start)/forcing_length)*(t-forcing_start)+section_start
+            return p0 +
+                   forcing_scale*(
+                p(time_shift)-p(section_start)
+            )/(p(section_end)-p(section_start))
```
Datseris, line 103: This shouldn't be computed all the time because iti s a fixed window. compute it once, and store it into a new field of `RateConfig` during creation.

src/r_tipping/RateSystem.jl
Datseris: Okay after going through the _whole_ file, I finally understant that this is `Δt`. Can we please use the terminology we agreed upon after 2 videocalls? This keyword (`Δt`) and the next (`Δp`) should be the only keywords given to this type.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+- `forcing_scale = 1.0`: Amplitude of the ramping. The ramping is then automatically rescaled 
+- `t0 = 0.0`: Initial time of the resulting non-autonomous system (relevant to later compute trajectories)
+
+## Description
+The returned `RateSystem.system` is is 
+autonomous before `forcing_start`, 
+non-autnonmous from `forcing_start` to `forcing_start+forcing_length` with the parameter shift given by the [`RateConfig`](@def), and again 
+autonomous after `forcing_start+forcing_length`.
+
+## Fields of the resulting `RateSystem` type
```
Datseris, line 54: Delete this section.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+    if forcing_start < t0
+        @warn "The forcing starts before the system initial time t0=$(t0)."
+    end
+
+    system = apply_ramping(
+        auto_sys, rc, pidx, p0, params, forcing_start, forcing_length, forcing_scale, t0
+    )
+    forcing = t -> p_modified(t, rc, p0, forcing_start, forcing_length, forcing_scale)
+
+    return RateSystem(system, forcing, pidx)
```
Datseris, line 81: I'm missing a function here. Two in fact. We need to API functions to change the forcing level and rate. We should define two functions, `set_forcing_size!(rs, Δp)` and `set_forcing_duration!(rs, Δt)` so that this system can be used downstream to e.g., find the critical forcing rate under constant `Δp`. Or, find the critical parameter increase under constant `Δt`. Etc.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+    return q
+end
+
+# returns a continuous time dynamical system with modified drift field
+function apply_ramping(
+    auto_sys, rc, pidx, p0, params, forcing_start, forcing_length, forcing_scale, t0
+)
+    f = deepcopy(dynamic_rule(auto_sys))
+
+    function f_new(u, p, t)
```
Datseris, line 118: In the email you said that this is 3x slower...? Wow. This is very bad! We should brainstorm a bit alternative ways to do this. Typically you'd hope to have a slowdown of 10-50% when going from a hard-coded to a general solution but here we are talking about a slowdown of 300% which is massive. @reykboerner can you also help tjhe brainstorming please? I'll also think of something.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+end
+
+"""
+    RateSystem(sys::ContinuousTimeDynamicalSystem, rc::RateConfig, pidx; kwargs...)
+
+Creates a `RateSystem` type from an autonomous dynamical system `sys` and a time-dependent
+parametric forcing protocol of `RateConfig` type.
+
+## Keyword arguments
+- `forcing_start = RateConfig.section_start`: Time when parameter shift starts (before this, the resulting system will be autonomous)
```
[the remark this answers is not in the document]
Datseris: Oooooh, no, now I finally understand. This isn't really related with the "section start/end". This is just `t0`.

src/r_tipping/RateSystem.jl
Datseris: Let's also add the `set_forcing_Start!(rs, t0)` for setting the time the forcing starts.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+The `RateConfig` type allows to specify the functional form of a parametric
+forcing over a time interval. This forcing protocol can then be applied to the parameter
+of a dynamical system using the `RateSystem` constructor, which also allows to modify
+the rate and magnitude of the forcing.
+"""
+mutable struct RateConfig
+    pfunc::Function
+    section_start::Float64
+    section_end::Float64
+end
```
Datseris, line 24: I'd like us to add here a couple of convenience functions, so that `RateConfig` can be called with a symbol:
`RateConfig(:linear)` makes a linear function, `RateConfig(:tanh)` makes the stadard tanh (with a slope of 1 at its midpoint). Later down the line we can add more convenience functions if need be.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,133 @@
...
+
```
Datseris, line 25: After this PR is merged, I will contribute a `reverse = false` option to the config. If true, the forcing is reversed in the timewindow `Δt, 2Δt` using exactly the same functional form used to scale it.

Datseris:
Oh, I've missed this:

```julia
nonauto_traj = trajectory(nonauto_sys.system, T, x0);
```
this is not the way to go. There is a well defined API for what a `DynamicalSystem` is, and this `RateSystem` should be **exactly** a dynamical system. Make it subtype `ContinuousTimeSystem` so that the call `trajectory(nonauot_sys)` works directly.

You guys have done this work already when you created the `CoupledSDEs` type. You already have all the skills and knowledge for this, no need to wait for me to comment on this, just go ahead and finalize things.

reykboerner:
Thanks @Datseris for your review! So apart from the computational speed, do you find the interface design makes sense?

> If you don't like `Δp, Δt`, then we can go with : `forcing_length` (Δt), `forcing_size` (Δp) and `forcing_start`.

I wasn't in the video calls, so I didn't know about these discussions about the naming. I prefer the written out labels because `Δt` looks like a time step and `Δp` is not general enough. I strongly think that we should not restrict to monotonic forcing protocols in the long-run. `forcing_scale` more accurately describes what this setting does: it scales the magnitude of the `RateConfig` you provided. Also `forcing_magnitude` works in my opinion if you prefer.


> this is not the way to go. There is a well defined API for what a `DynamicalSystem` is, and this `RateSystem` should be **exactly** a dynamical system. Make it subtype `ContinuousTimeSystem` so that the call `trajectory(nonauot_sys)` works directly.

Happy to do this. One question: Is it a problem that the `RateSystem` type itself contains a `CoupledODEs` type as `RateSystem.system`? Because essentially the `RateSystem` is just a container for the nonautonomous system + the forcing config. There is an equivalence between a dynamical system constructed via `RateSystem` vs. the same system constructed as an explicitly time-dependent `CoupledODEs`.

Datseris:
`forcing_scale` is fine as well and I agree with you with nonmonotonic forcings. I stated in one of the comments i'll contribute reversible forcing used to study overshooting (doing this right now for the lectures, might as well put it here)

---

> Happy to do this. One question: Is it a problem that the `RateSystem` type itself contains a `CoupledODEs` type as `RateSystem.system`?

no, this is no problem at all. Essentially make sure that you subtype `RateSystem <: ContinuousTimeDynamicalSystem` and then extend the functions to dispatch to the field, a-la how you did four copled sdes 
```julia
for f in (:step, :trajectory, ....)
    @eval DynamicalSystemsBase.$(f)(rs::RateSystem, args...; kw...)= $(f)(rs.system, args...; kw...)
end
```

Datseris:
> So apart from the computational speed, do you find the interface design makes sense?

yes provided the changes I've outlined above are included!

oameye:
> this is absolutely massive, and the number of allocations shows very clearl problems in the implementation beyond the algorithm design. @raphael-roemer have you profiled your code and checked the basic components of julia optimization, (such as type instability, unecessary allocations, etc.), as e.g., mentioned here: [modernjuliaworkflows.org/optimizing](https://modernjuliaworkflows.org/optimizing/) ?

@raphael-roemer Hey, if needed I am happy to jump in here and to optimize the code. Just let me know.

src/r_tipping/RateSystem.jl
```diff
@@ -1,12 +1,18 @@
+@kwdef mutable struct RateSystemForcing
+    pidx::Int
+    rc::RateConfig
+    t0::Real
+    forcing_start::Real
```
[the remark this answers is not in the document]
oameye: Maybe just add `RateSystemForcing` to `RateSystem` directly :P

oameye:
A preview of the documentation can be seen here:
https://juliadynamics.github.io/CriticalTransitions.jl/previews/PR117/

src/r_tipping/RateSystem.jl
oameye: Why did you resolve this @reykboerner. I agree with Datseris that `Δt` and `Δp` would be much better.

src/r_tipping/RateSystem.jl
oameye: These are added now

src/r_tipping/RateSystem.jl
oameye: The docstring has been made better

oameye:
I think all comments should be addressed now. Let me know if I forgot something.

@Datseris I am sorry for getting ahead of myself yesterday. You are right the PR was not ready for merge.

src/r_tipping/RateFunction.jl
```diff
@@ -0,0 +1,54 @@
+"""
+    RateFunction(;pfunc::Function, ptspan)
+
+Time-dependent forcing protocol ``p(t)`` describing the evolution of a parameter over a
+time interval `ptspan = (start, end)`. Used to construct a non-autonomous `RateSystem`.
```
reykboerner: The idea here was that `p(t)` is first defined in arbitrary units - only when defining the RateSystem the forcing function is scaled in the system's time units. That's why I would write "domain interval" instead of "time interval"

src/r_tipping/RateFunction.jl
```diff
@@ -0,0 +1,54 @@
+"""
+    RateFunction(;pfunc::Function, ptspan)
```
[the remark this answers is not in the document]
Datseris: `ForcingProfile` better reflects what this does. This part of the code cannot be named "rate" as it doesn't involve time.

src/r_tipping/RateSystem.jl
```diff
@@ -0,0 +1,169 @@
...
+            setproperty!(p, pidx, pvalue)
+        end
+        return f(u, p, t)
+    end
+    old_prob = deepcopy(referrenced_sciml_prob(ds))
+    prob = remake(old_prob; f=f_new, p=params, tspan=(t0, Inf))
+    return CoupledODEs(prob, ds.diffeq)
+end
+
+function set_forcing_scale!(rs::RateSystem, scale) end
```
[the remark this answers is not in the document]
Datseris: what needs to be written?

reykboerner:
Hey @oameye @raphael-roemer @ryandeeley I just had a quick call with @Datseris to make the `RateSystem` code more flexible and compatible with the `DynamicalSystem`s API - please don't revise this PR further until I push the changes we drafted so we don't double the work

Datseris:
> The main change is a mutable struct `NonautonomousDynamicRule` which provides a [function-like object](https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects) that stores all the info needed to construct/modify a `RateSystem`.

Do note that the benchmarks have to be run again with this change, to see if we have a serious regression or not.

Datseris:
i am sorry, i only now finished work so as you can imagine what i said above was completely bs and a false promise. i "promise" i will look at it on monday night.

Datseris:
I've listed some todos in the source code. One of the todos is to change the parameter so that we move into the [index => value] syntax for multiple parameters. This can be done in a different PR.

reykboerner:
The `RateSystem` version still seems faster than the hard-coded one, interestingly:

```
"Hard coded" => TrialEstimate(45.375 μs)
"RateSystem" => TrialEstimate(28.000 μs)
```

oameye:
On CI it is [less severe](https://github.com/JuliaDynamics/CriticalTransitions.jl/actions/runs/19048113149/job/54401340450?pr=117#step:5:551). However, I also think this can't be real. We should try out a larger/complex system.

PR #139, opened 2024-12-26
docs: add docstring for MaximumLikelihoodPath struct

oameye:


src/largedeviations/MaximumLikelihoodPath.jl
```diff
@@ -1,3 +1,18 @@
...
+- `path::Matrix{T}`: The path matrix.
+- `action::T`: The action value associated to the path.
+- `path_history::Phis`: The history of paths in the optimisation algorithm (optional).
+- `action_history::Ahis`: The history of action of the paths in the ptimisation algorithm (optional).
+- `λ::Lambda`: The Lagrange multiplier parameter for the maximum likelihood path.
+- `generalized_momentum::GPV`: The generalized momentum of the phase space variables (optional).
+- `path_velocity::PV`: The path velocity (optional).
+
+"""
 mutable struct MaximumLikelihoodPath{T,Phis,Ahis,Lambda,PV,GPV}
```
reykboerner, line 16: Is "maximum likelihood path" the common term in your field? If we are only considering methods based on minimizing a cost functional, I would prefer `MinimumActionPath`.

Also, we could let `path` be of type `StateSpaceSet`, what do you think?

oameye: That's okay for me. Also for the StateSpaceSet, but make sure that the type is concrete. That's better for performance

oameye: Feel free to open a PR. I can also do it later

oameye: In the original paper they call it [Maximum Likehood Path](https://doi.org/10.1103/PhysRevLett.100.140601). In my field they call it Instanton.

PR #141, opened 2024-12-27
refactor transitions

oameye:


oameye:
@reykboerner could you take a look at PR. Feel free to change some things.

src/trajectories/transition.jl
```diff
@@ -110,55 +121,52 @@ See also [`transition`](@ref).
...
 """
+
 function transitions(
     sys::CoupledSDEs,
-    x_i,
-    x_f,
-    N=1;
-    rad_i=0.1,
-    rad_f=0.1,
+    x::NTuple{2},
```
reykboerner, line 127: What is the reason for switching from `x_i`, `x_f` to a tuple `x`? I find the former much more intuitive since there is a direction associated with the transition

oameye: Ha yeah, I liked it because it was more concise. I am not a fan of too many `args` and `kwargs` in function. But I don't have strong opinion, if you want it reversed, I'll do it :)

oameye: Are you okay with adding both methods?

test/trajectories/transition.jl
```diff
@@ -11,17 +11,17 @@
...
     @test succes
     @test time[end] < 1e3
-    @test norm(trajectory[:, 1] - fp1) < 0.1
-    @test norm(trajectory[:, end] - fp2) < 0.1
+    @test norm(tr[1, :] - fp1) < 0.1
+    @test norm(tr[end, :] - fp2) < 0.1
 
     ensemble = transitions(sys, fp1, fp2, 10)
     @test ensemble.success_rate ≈ 1.0
     @test ensemble.t_trans ≈ 4.493941793363376 atol = 1e-2
```
reykboerner: Doesn't this give very different numbers each time? `atol=1e-2` seems very small

src/sde_utils.jl
```diff
@@ -44,3 +44,5 @@ dependent systems, the time can be specified as a keyword argument `t` (by defau
 function div_drift(sys::CoupledSDEs, x, t=0)
     return tr(jacobian(sys)(x, sys.p0, t))
 end;
+
+trajectory_algorithm(ds::CoupledSDEs) = ds.integ.alg
```
reykboerner, line 53: Should we just call this function `solver(ds::CoupledSDEs)` for simplicity?

src/trajectories/transition.jl
```diff
@@ -1,9 +1,10 @@
-struct TransitionPathEnsemble
-    paths::Vector
-    times::Vector
-    success_rate::Real
-    t_res::Real
-    t_trans::Real
+struct TransitionPathEnsemble{SSS,T,Tstat,ES}
```
reykboerner, line 1: We could rename this to `TransitionEnsemble` as you suggested, it's indeed shorter

reykboerner (commented):
I made a few tiny changes in my last commit, plus the comments of this review, otherwise I think it looks great!

src/trajectories/transition.jl
reykboerner: I agree that we shouldn't overload functions but in this case I prefer having the arguments separately as I find it more clear. Having both methods seems a bit overkill if the only difference is the input as tuple vs. two args?

oameye:
@reykboerner Go for it 😊

test/trajectories/transition.jl
oameye: Normally with the Seed it should give the same every time

src/sde_utils.jl
oameye: Yeah, sounds good

src/trajectories/transition.jl
oameye: Yes I agree

test/trajectories/transition.jl
reykboerner: Ah, ok. I thought seed didn't work on github actions? This now resolved?

oameye: It seems to work now on CI 😊

PR #148, opened 2024-12-28
feat: make the string method compatiable with StateSpaceSet

oameye:


Datseris:
What are the "operations"? Are they Linear algebra, which require data inherently in a mtrix form? Or are they vector algebra, which operate directly on vectors, and sets of vectors?

oameye:
It just how it is implemented. It aim for non-allocating code by having a iterative, mutation-heavy numerical kernel where the the workspace consists largely of dense matrices. Its just the the mathematical organization of the algorithm.
```julia
path .= ws.update
copyto!(path_prev, path)
interpolate_path!(path, ...)
view(path, :, i)
similar(path)
```

I agree we could rewrite it for `StateSpaceSet`, but I also want to support high dimensional systems. So I need  `StateSpaceSet` to have a proper interface where it support both presentations.

Datseris:
It does already, as you can have the inner vectors represented as `Vector` instead of `SVector`.

PR #155, opened 2025-01-01
feat: EnsembleStatistics

### LINKED ISSUE #1: New types for simulation data
Currently the main structure of CriticalTransitions.jl is the `StochSystem`, and the typical syntax for methods is `function(sys::StochSystem, args; kwargs...)`.

A typical workflow for numerically studying critical transitions could look like this:
1. Investigate system properties (fixed points, stability, basins of attraction, flow field, quasipotential, instanton, ...)
2. Generate sample trajectories (direct simulation, rare event/importance sampling, pathspace Langevin MCMC sampling, ...)
3. Analyze the statistical properties of samples (mean transition path, residence time distribution/scaling, invariant measure, ...)
4. Relate these sample properties to the dynamical system properties (boundary crossing point, transition indicators, quasipotential height differences, ...)
5. (what else would be of interest?)

The `StochSystem` structure is convenient for points 1 and 2 of this workflow. To deal with sampling data in a unified and practical way, it could be useful to define new structures:
* `Trajectory` and `TrajectoryEnsemble`
* `Transition` and `TransitionEnsemble` (subtypes of `Trajectory` and `TrajectoryEnsemble`, respectively)

These structures could be abstract types with fields. For example, `Transition.path` could output the path as a matrix of size `(D x N)`, where `D` is the system dimension and `N` the number of path points.

Functions for analyzing transition samples could then follow the general syntax `function(tr::TransitionEnsemble, args; kwargs...)` or `function(sys::StochSystem, tr::TransitionEnsemble, args; kwargs...)`.

Would this be a good way to structure the functionality?

oameye:
resolve #1

### DIFF
--- a/docs/src/man/largedeviations.md
+++ b/docs/src/man/largedeviations.md
@@ -3,7 +3,7 @@
 
 ## Mimumum action path
 ```@docs
-MinimumActionPath
+CriticalTransitions.MinimumActionPath
 ```
 
 ## String method
--- a/docs/src/man/sampling.md
+++ b/docs/src/man/sampling.md
@@ -1,7 +1,8 @@
 # Sampling transitions
 
 ```@docs
-TransitionEnsemble
+CriticalTransitions.TransitionEnsemble
+CriticalTransitions.TransitionStatistics
 ```
 
 ## ... by direct simulation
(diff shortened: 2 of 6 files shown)


reykboerner (approved):
Cool, looks good to me! If the docs build, good to merge from my side :)

PR #178, opened 2025-06-21
docs: overhaul

oameye:


reykboerner:
Some thoughts regarding [TPT](https://juliadynamics.github.io/CriticalTransitions.jl/dev/man/tpt/) and sgMAM @oameye 

1. It seems we need two extra types besides `CoupledSDEs`, namely:
     - `ExtendedHamiltonian` (similar to what is now called `SgmamSystem` - is there a name for it in the literature?)
     - `LangevinSystem` (what is now called `Langevin`)
2. Mathematically, `Langevin` and `CoupledSDEs` are both ways to write a Langevin equation, so we should clarify their connection in the docs. I also wonder whether `Langevin` could be formulated more similarly to `CoupledSDEs` to make the similarity clear

What do you think and would you have some time to work on this @oameye ?

oameye:
I am happy to work on it. I am trying something else by Thursday and then I will try to draft something up and try to address the questions.

In the end I think the TPT should go to some experimental section

reykboerner:
> I am happy to work on it. I am trying to finish something else by Thursday and then I will try to draft something up and try to address the questions.

Great, see new issue for this @oameye (#185)

PR #182, opened 2025-07-28
Package description paper

reykboerner:
This branch will be used to write and submit the paper describing v1.0 to JOSS (see https://joss.readthedocs.io/en/latest/submitting.html).


ryandeeley:
Hello, I will perhaps directly push these minor changes at the end of the day but am just listing them here as I go for now:

Title page section:

- I would add "nonautonomous dynamics" to the tags if there is room.

Summary section: 

- Possibly change “a key element of metastable behavior are…” to “a key association with metastable behavior is …”. For me this sounds a little like you are trying to define metastability with respect to critical transitions, although the meaning of metastability is not unanimously accepted. I personally don’t think you need critical transitions to say that metastability is present in a system, but rather that the presence of metastability becomes relevant once you can attain an alternative (metastable) state via a critical transition through external forcing.

Statement of need section:

- I stumbled a little over the sentence “Research problems are often not tractable analytically, requiring numerical techniques for applying dynamical systems theory.” Maybe I would say: “Research problems are often not tractable analytically and (rather) require numerical techniques to apply concepts from dynamical systems theory.”   
- I would change “proposed” to “developed”.
- Change “had not been implemented” to “have not been implemented”.
- I personally would remove the commas around sometimes and/or change to “in some cases” - I think “in some cases” adds clarity to what I think you mean by this (in the case of some algorithms) and I think the commas are unnecessary and cause people to slip up during the sentence and lose the meaning.  

I very much like the way you've introduced the utilities of `SciML.jl` and `DifferentialEquations.jl` and said that `DynamicalSystems.jl` applies these to create an ecosystem for analysing autonomous, deterministic dynamical systems, and then proceeded to say that a similar ecosystem for forced systems is missing. I think this makes very clear why there is a "need" for `CriticalTransitions.jl` and what it provides!

PR #188, opened 2025-08-02
refactor: make tpt experimental

oameye:


src/experimental/transition_path_theory/langevin.jl
```diff
@@ -10,7 +10,7 @@ $(TYPEDFIELDS)
 $(METHODLIST)
```
reykboerner, line 11: Would be nice to add the Langevin equations of motion here in the docstring to see how it relates to the fields of the struct

I'll review the rest by Monday!

reykboerner (approved):
If you agree with my edits, I think it's all good to go! Just please regenerate the example .md file with Literate.jl

Thanks for getting this in shape @oameye :)

src/experimental/transition_path_theory/langevin.jl
oameye: Thanks Reyk for cleaning up the example. Appreciated it!

PR #267, opened 2026-01-29
refactor and rename: minimum action methods with type based optimizers 

oameye:
Move the optimization algorithm selection from keyword arguments to a type-based dispatch system. Instead of passing `optimizer` and `stepsize` as separate kwargs, callers now pass a typed optimizer object (e.g. `GeometricGradient(; stepsize=1e3))` as a positional argument. This applies to `geometric_min_action_method`, `simple_geometric_min_action_method`, and `min_action_method`.

This enables dispatching on optimizer type, which is the foundation for adding a stepsize controller in #268.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ examples/Manifest.toml
 benchmarks_output.json
 benchmarks/Manifest.toml
 Manifest.toml
+.claude/*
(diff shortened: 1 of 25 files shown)


reykboerner (changes requested):
We could use this moment to rename the functions by removing the `_method` suffix, as we once discussed. Thus the new names would be
- `min_action` for MAM
- `geometric_min_action` for gMAM
- `simple_geometric_min_action` for sgMAM

What do you think?

PR #268, opened 2026-01-29
feat: backtracking step-size control for GeometricGradient

oameye:
Adds an Armijo-style backtracking line search to the GeometricGradient optimizer used in `minimize_simple_geometric_action` . Instead of requiring users to hand-tune a fixed step size (too small = slow, too large = diverges), the optimizer now automatically shrinks the step when needed, giving both speed and accuracy.

  Changes:
  -`  GeometricGradient` now defaults to max_backtracks=10; pass max_backtracks=0 to restore the old fixed-step behavior.
  - Default stepsize raised from 1e-1 to 1e3 to match the new backtracking default.
  - Removes the `DataStructures` dependency (replaced a 2-element circular buffer with a simpler implementation).
  - Adds a new example (backtracking_KPO.jl) demonstrating the accuracy/speed tradeoff on the Kerr parametric oscillator.

needs #267 

https://juliadynamics.github.io/CriticalTransitions.jl/previews/PR268/


examples/backtracking_KPO.jl
```diff
@@ -0,0 +1,216 @@
...
+lines!(
+    ax,
+    res_bt.path[:, 1],
+    res_bt.path[:, 2];
+    label="backtracking (S=$(round(res_bt.action, sigdigits=5)))",
+    linewidth=2,
+    color=:red,
+)
+axislegend(ax; position=:lt)
+fig_paths
```
Datseris: I'd recommend reducing the amount of breaklines used. You don't want the user to be scrolling forever until they encounter an output, especially for lines of code so unrelated to the package (plotting)

oameye: agreed

oameye:
@reykboerner this is ready to merge. Check https://juliadynamics.github.io/CriticalTransitions.jl/previews/PR268/examples/backtracking_KPO/ for the results

PR #280, opened 2026-03-08
Towards merging into DynamicalSystems.jl

Datseris:
- Create a central Tutorial following the exceptional documentation tutorial https://www.youtube.com/watch?v=ZDQ28X3mGSo
- Remove extensions whose only purpose is to convert SDEs to ODEs: this incredibly simple operation can be done by the user, thereby allowing us to remove 5 extra files and circular dependencies
- Move the docstyle to JuliaDynamics
- Working into making the docs more similar to rest of DynamicalSystems.jl with a central API page, and examples pages (examples already exist)

Note: this PR made me realize that the `RateSystem` is far from being done. In the rate system files there are at least 6-8 TODOs all with substantial weight. 

Note 2: removing the extensions removed a ChaosTools.jl additional method for specifying the boxes as vectors of mins and vectors of maxes. In my opinion such trivial operations done in one line by the user should not warrant making a whole extension. I argue in fact they should not warrant providing a conversion function to begin with.

What remains is to add some example applications for both random and rate systems.

Note3: it is worth introducing a generic supertype `RandomSystem`. Its only child at the moment is `CoupledSDEs` but in the future there may be more types allowable. I know a research group at imperial works a lot with random systems (primarily maps) with finite noise, which have completely different behaviour than wiener process SDEs.


Datseris:
Not much, I am basically merging the getting started and tutorial pages into a new tutorial which covers both random and rate systems. The examples stay as they are. A new API page is necessary which lists all functions exported by the module. I haven't exactly understood where you guys do this at the moment.

oameye:
could we maybe merge this in a dev branch. Would be nicer to have CI working on main :p

PS: thesis deadline is by the end of march. From then one, I hope to push this package to completion.

PR #316, opened 2026-05-04
Finalise RateSystem

raphael-roemer:
Implemented all TODOs from our meeting.

src/RateSystem.jl:  
Now, the forcing profiles and their parameters must be given as dicts and ramping multiple parameters can be handled.
Should work for both CoupledSDEs and CoupledODEs.
The docstrings are all updated as well.

test/r_tipping/RateSystem.jl: 
added a new tests, including a short compatibility test with ModelingToolkits and both for CoupledSDEs and CoupledODEs.


raphael-roemer:
TBD: 

- [ ] 1) update docs to be consistent with the new dict-based RateSystem definition and add CoupledSDEs example
- [x] 2) add more and better testsets for the new RateSystem setup
- [x] 3) make `RateSystem` work for input CoupledODEs and CoupledSDEs, in-place and out-of-place

benchmarks/ratesystem.jl
```diff
@@ -22,7 +22,7 @@ function benchmark_rate_system!(SUITE)
     forcing_scale = 1.0
     t0 = forcing_start_time
 
-    rs = RateSystem(ds, fp, pidx; forcing_start_time, forcing_duration, forcing_scale, t0)
+    rs = RateSystem(ds, Dict(pidx => fp); forcing_start_time = forcing_start_time, forcing_duration = forcing_duration, forcing_scale = forcing_scale, t0 = t0)
```
reykboerner, line 25: Note: `kwarg = kwarg` can be simplified to `kwarg` if it is called the same

examples/tutorial.jl
```diff
@@ -64,7 +64,11 @@ fp = ForcingProfile(:linear)
 # The remaining information is encoded when creating the `RateSystem`:
 
 pidx = 6 # which parameter changes
-rs = RateSystem(ds, fp, pidx; forcing_start_time = 10, forcing_duration = 10, forcing_scale = 5)
+rs = RateSystem(ds, Dict(pidx => fp);
```
reykboerner: I added a convenience constructor that allows to use the same syntax as before for 1-parameter cases

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    the out-of-place call and copies the result into `du`.
+
+The `RateSystemSpecs` ensures parameter values are adjusted according to the
+configured forcing profiles before delegating to the underlying autonomous rule.
+"""
+mutable struct RateSystemSpecs{R,K,T,PC,E} <: Function
+    "Dynamic rule of the underlying autonomous system"
+    unforced_rule::R
+    "Mapping parameter index => ForcingProfile"
+    forcing_profile::Dict{K,ForcingProfile}
```
reykboerner: Renamed this to `forcing_profile` (instead of `forcers`) to match the rest of the naming and be more specific

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    "Mapping parameter index => forcing scale"
+    forcing_scale::Dict{K,T}
+    "Mapping parameter index => initial (autonomous) parameter value"
+    p0::Dict{K,E}
+    "Initial time (of system initiation)"
+    t0::T
+    "Dummy container (copy of parameters)"
+    pdummy::PC
+    "Reference to the owning (Coupled) system, set after construction"
+    owner::Any
```
reykboerner: What is the`owner` field for and why is it called like that?

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+end
+
+# Accessors for initial parameter(s). These are explicit and avoid overriding `getproperty`.
+"""
+    initial_parameter_map(rss::RateSystemSpecs)
+
+Return the internal mapping (`Dict`) of parameter keys to the initial (autonomous)
+parameter values stored in `rss.p0`.
+"""
+function initial_parameter_map(rss::RateSystemSpecs)
```
reykboerner: why are these three functions needed? Can one not do `parameters(rs, t0)`?

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+# TODO: this must be rewritten using `set_parameter!` or its source code.
+# otherwise it doesn't work with ModelingToolkit.jl;
+# Or better yet, use `set_parameters!` and give a dict of parameters to set?
+function (rss::RateSystemSpecs)(u, p, t)
+    pmod = p_modified(rss, p, t)
+    # Prefer calling out-of-place signature `f(u,p,t)` which returns du
+    return rss.unforced_rule(u, pmod, t)
+end
+
+function (rss::RateSystemSpecs)(du, u, p, t)
```
reykboerner, line 151: Todo: need to add a test for in-place systems

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
 For multiple parameters, if `pidx` is given, change the forcing only corresponding to
 the specified parameter, otherwise update the forcing scale of _all_ parameters to `scale.`
-"""
-function set_forcing_scale!(rs::RateSystem, scale)
-    rs.forcing.forcing_scale = scale
-    return rs
-end
-
+"""
+function set_forcing_scale!(rs::RateSystem, scale; pidx=nothing)
```
reykboerner: make `pidx` an optional positional argument instead of a keyword argument?

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    # semantics when we would need to mutate solver-owned parameter containers.
+    try
+        # If the wrapped system isn't in-place, we can't safely advertise it.
+        if !SciMLBase.isinplace(rs.system)
+            return false
+        end
+
+        # Obtain concrete argument types for the current system state/params.
+        u = DynamicalSystemsBase.current_state(rs)
+        p = DynamicalSystemsBase.current_parameters(rs)
```
reykboerner: is there a time argument `t` missing here?

test/r_tipping/RateSystem.jl
```diff
@@ -88,3 +260,27 @@ end
         @test all(tr_constructed .≈ tr_hardcoded)
     end
 end
+
+@testset "RateSystem SDE smoke" begin
```
reykboerner: Smoke?

raphael-roemer: "Smoke" because it is just a very short test that only checks whether the system builds correctly. We still have to expand this test

src/r_tipping/RateSystem.jl
raphael-roemer: These both work without `t`, see here: https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/v3.15.2/src/core/dynamicalsystem_interface.jl#L185-L190
https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/v3.15.2/src/core/dynamicalsystem_interface.jl#L242-L249

src/r_tipping/RateSystem.jl
raphael-roemer: It should be possibe to do something like `parameters(rs, minimum(values(rs.forcing.forcing_start_time))` instead.  However, it may in some cases be convenient to have the option to access the unforced parameters directly from the `RateSystemSpec` instead of only having a function to get it from the `RateSystem`. 
But I don't have strong feelings about this - happy to change this

src/r_tipping/RateSystem.jl
raphael-roemer: `owner` is used to allow a fallback option in case the parameter container is not a `Dict` or a `Vector` (which might be the case for some `ModelingToolkit` functionality)
Then, using `owner`, the `p_modified()` function can try to directly modify the parameter container of the underlying `CoupledSDEs` or `CoupledODEs`

src/r_tipping/RateSystem.jl
raphael-roemer: I think that's a good idea. Should we still keep the keyword argument to not break users' code, or just fully switch to the optional positional argument?

src/r_tipping/RateSystem.jl
raphael-roemer: Do you mean that there should be another testset in `test/r_tipping/RateSystem.jl` for in-place systems?

src/r_tipping/RateSystem.jl
Datseris: But why is the type `Any`?

src/r_tipping/RateSystem.jl
Datseris: I think we should remove this function and aim for an initially minimal api; indeed, doing `parameters(rs, t0)` should be what we offer for now. Remember that `t0`is a user input, as such user always has access to it, they don't have to do `rs.forcing....`.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+
+Return the internal mapping (`Dict`) of parameter keys to the initial (autonomous)
+parameter values stored in `rss.p0`.
+"""
+function initial_parameter_map(rss::RateSystemSpecs)
+    return getfield(rss, :p0)
+end
+
+"""
+    initial_parameter(rss::RateSystemSpecs, pkey)
```
Datseris: same comment as above

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    end
+end
+
+# RateSystem-level wrappers:
+"""
+    parameters(sys, args...; kw...)
+
+Convenience wrapper delegating to `current_parameters(sys, ...)` for API
+compatibility with earlier code and callers expecting a `parameters` helper.
+"""
```
Datseris: There is no need for compatibility with earlier code as we are going for a breaking release. We don't have to spend effort on that or maintain backwards compatible code at this stage of the project (in my opinion).

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+Description
+The profile `interval` defines the domain of the forcing function; when applied
+to the system the profile is rescaled in system time units using the
+configured `start` and `duration` values - this allow changing the rate of the 
+parameter ramping. Before a forcing_profile's `start` time the parameter equals its initial 
+autonomous value; during the forcing interval it follows the rescaled profile (multiplied 
+by the corresponding `forcing_scale` factor); after the interval the parameter is frozen 
+at its final value.
+
+Multiple parameters
```
Datseris: header `##` missing here

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
-# TODO: ensure all three key update functions follow the optional [, pidx] syntax
-# for multi parameters.
+"""
+        RateSystem <: ContinuousTimeDynamicalSystem
+        RateSystem(ds::ContinuousTimeDynamicalSystem, forcing_profile::Dict; kw...)
+
+Construct a non-autonomous dynamical system by applying time-dependent parametric
+forcing protocols to an underlying autonomous continuous-time dynamical system `ds`.
+
+`forcing_profile` must be a `Dict` mapping parameter indices (anything valid for
```
Datseris: we can provide a convenience constructor that is identical with the constructor of the previous version so that this does not ""must" be the case, but "can" be.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    "Non-autonomous continuous-time dynamical system"
+    system::S
+    "Data structure representing the system and forcing specifications in system units"
+    forcing::RateSystemSpecs{R}
+end
+
+# RateSystem-level wrappers (placed after `RateSystem` type definition)
+function initial_parameter_map(rs::RateSystem)
+    return initial_parameter_map(rs.forcing)
+end
```
Datseris: this function should be 1 line of code, same for all the rest.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+
+function initial_parameter_value(rs::RateSystem)
+    return initial_parameter_value(rs.forcing)
+end
+
+"""
+    RateSystem(ds::ContinuousTimeDynamicalSystem, forcing_profile::Dict; kwargs...)
+
+Constructor of a [RateSystem](@ref) with one or more time-dependent forcing parameters.
+"""
```
Datseris: what's the difference of this docstring with the above? Perjhajps this docstring can be removed?

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    # set owner and ensure pdummy has same shape as system parameters
+    rss.owner = newds
+    rss.pdummy = deepcopy(current_parameters(newds))
+    return RateSystem(newds, rss)
+end
+
+"""
+    RateSystem(ds::ContinuousTimeDynamicalSystem, forcing_profile::ForcingProfile, pidx; kwargs...)
+
+Constructor for a [`RateSystem`](@ref) with one time-dependent parameter.
```
Datseris: same. we can discuss the ways to construct the rate system in one docstring instead of having different ones, i believe this is easier to navigate because if ou discuss one way of construction it is easy to discuss all of them at the same place for the user.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+            if isa(pd, AbstractDict)
+                pd[pkey] = pt
+            elseif isa(pd, AbstractVector)
+                pd[Int(pkey)] = pt
+            else
+                # Try per-parameter setter on owning system if available
+                if rss.owner !== nothing
+                    try
+                        DynamicalSystemsBase.set_parameter!(rss.owner, pkey, pt)
+                    catch
```
Datseris: oh, try catch blocks are absolutely fatal for performance. And here the try catch is in the inner most possible loop. This has to be reworked. 

Why is this function so complex to begin with, wjhy aren't we just calling at the top level the `DynamicalSystemsBase.set_parameter!`function?

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+            else
+                pt = p0 + scale * (f(section_end) - f(section_start))
+            end
+
+            p[idx] = pt
+        end
+        return p
+    else
+        # Arbitrary parameter container: build a copy in pd and try to apply updates
+        pd = deepcopy(rss.pdummy)
```
Datseris: definitely no need to be deep copying parameters in the innermost loop. Remember, this function is called 10s of times at a single `step!` call. So this particualr code line is called typically millions to billions times during a single `trajectory` generation.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+- For arbitrary parameter container types, a copy of `rss.pdummy` is created and
+    updated. When `rss.owner` is available, `DynamicalSystemsBase.set_parameter!`
+    (per-key) or `set_parameters!` (bulk) is used where supported; otherwise
+    `setproperty!` is attempted on the copy.
+
+The returned container has the same general shape/type expected by the
+underlying system.
+"""
+function p_modified(rss::RateSystemSpecs, p, t::Real)
+        # In-place update for Dict and Vector parameter containers (mutates `p`),
```
Datseris: this whole function needs to be reworked to call `DDynamicalSystemsPBase.set_parameter!` at the top level, we really shouldn't be duplicating shource code here, especially since the base function is already specialized to work with all types of parameters compatibles with the ecosystem.

src/r_tipping/RateSystem.jl
Datseris: There is no user code to break, this is not released functionality. it's brand new.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+) # all api functions here
+    @eval DynamicalSystemsBase.$(f)(rs::RateSystem, args...; kw...) = $(f)(
+        rs.system, args...; kw...
+    )
+end
+
+reinit!(rs::RateSystem, u=initial_state(rs); kw...) = reinit!(rs.system, u; kw...)
+
+SciMLBase.step!(rs::RateSystem, args...) = (SciMLBase.step!(rs.system, args...); rs)
+function SciMLBase.isinplace(rs::RateSystem)
```
Datseris: I don't understand why this function is so complicated. What was the problem wit hthe original function? If the original unforced suystem is inplace, so is the rate system. Just propagate the function to the referrenced unforced system.

src/r_tipping/RateSystem.jl
Datseris: Can we please try the suggestion i had in my comments? First, update once the `rss.pdummy`. There is no need to be three times in the code the copy of the pt = p0 + scale.... .

Then, call `set_parameters!` on the dummy.

src/r_tipping/RateSystem.jl
reykboerner: Okay, we can solve this with `set_parameters` and get rid of owner

src/r_tipping/RateSystem.jl
Datseris: Let's try  the design of `p_modified!` in the main vbranch, it is almost finished.

src/r_tipping/RateSystem.jl
```diff
@@ -17,289 +17,660 @@ Note: The units of ``t`` are arbitrary; the forcing profile is rescaled to syste
...
+    pmod = p_modified(rss, p, t)
+    # Prefer calling out-of-place signature `f(u,p,t)` which returns du
+    return rss.unforced_rule(u, pmod, t)
+end
+
+function (rss::RateSystemSpecs)(du, u, p, t)
+    pmod = p_modified(rss, p, t)
+    # Try calling in-place signature `f(du,u,p,t)` first; if not defined,
+    # fall back to out-of-place `f(u,p,t)` and copy into `du`.
+    try
```
Datseris: We must never use `try` blocks in performance revelant code. There is also no reason to "check" internally if the system is in place or out of place. It cannot be both. Icf the original system is in place, so is the rate. There is no alternative. So the inplace function must also be 4 lines of code, like the out of place.

Datseris: thats all the code you need for hte inplace function:


    pmod = p_modified(rss, p, t)
    rss.unforced_rule(du, u, pmod, t)
    return nothing

Datseris:
@raphael-roemer sorry if comments are mean i was having abad day!!!!!!!

reykboerner:
Working on this now based on the discussion with @Datseris at EGU

src/r_tipping/RateSystem.jl
reykboerner: I would fully switch to have only one call signature.

src/r_tipping/RateSystem.jl
reykboerner: The `current_parameter` function of DynamicalSystems.jl indeed works without `t`, but this gives the parameters only at the current system time. In a `RateSystem`, you'd want to know the parameter container at any given time, and therefore we need a new function `parameters(rs, t)` as was previously implemented in `v0.8.0`. I suppose one could multiple-dispatch `current_parameters(rs, t)` but to avoid confusion I would just add `parameters` to the API.

test/r_tipping/RateSystem.jl
reykboerner: Okay :)

Datseris:
i have some free time tonight and nothing to do so i'll push some code here

raphael-roemer:
Thank you for your thorough reviews and the good comments, George and Reyk:) I see that you already started implementing them as well, thanks!
I will have time to work on further integrating the remaining comments on Wednesday evening. Please let me know if this interferes with your plans, so that we can circumvent difficult merge conflicts.

src/r_tipping/ForcingProfile.jl
```diff
@@ -0,0 +1,46 @@
...
+        return ForcingProfile(x -> x, (0.0, 1.0))
+    elseif sym == :tanh
+        return ForcingProfile(x -> tanh(x)/2 + 0.5, (-3.0, 3.0))
+    else
+        error("Only :linear or :tanh are supported input arguments.")
+    end
+end
+
+"""
+    data(fp::ForcingProfile; N=100)
```
Datseris: this is an incredibly generic name that we should not export; it hogs so many different names a user or other packages could be using...

Besides, do we actually need this function? Its source code is trivial enough to be done on the user side.

src/r_tipping/RateSystem.jl
```diff
@@ -1,305 +1,503 @@
...
+    system::S
+    "Data structure representing the system and forcing specifications in system units"
+    forcing::RateSystemSpecs{R}
+end
 
-## Description
+function RateSystem(
+    ds::ContinuousTimeDynamicalSystem,
+    forcing_profile::Dict;
+    forcing_start_time=nothing,
```
Datseris: forcing_start_time, duration, and scale, are not actually used anywhere inside this function body.

src/r_tipping/RateSystem.jl
```diff
@@ -1,305 +1,503 @@
...
-underlying autonomous system `ds`. At the end of the forcing interval
-(`t = forcing_start_time + forcing_duration`), the parameters are frozen at their final
-value and thereafter the system is again autonomous.
+    pdummy = deepcopy(current_parameters(ds))
 
-## Multiple parameters
+    rss = RateSystemSpecs{R,K,T,P}(
+        dynamic_rule(ds),
+        forcing_profile,
+        start_map,
```
Datseris: This is not defined anywhere

Datseris (commented):
I started reviewing this but I assume it is not finished, right? E.g the code won't run due to undefined variables.

src/r_tipping/ForcingProfile.jl
reykboerner: Ok, I'm removing it.

reykboerner:
> I started reviewing this but I assume it is not finished, right? E.g the code won't run due to undefined variables.

Yes, not finished yet.

src/r_tipping/RateSystem.jl
reykboerner: Done

src/r_tipping/RateSystem.jl
```diff
@@ -1,305 +1,499 @@
...
+        forcing_duration,
+        forcing_scale,
+        p0,
+        p0,
+        t0
+    )
 
-## Multiple parameters
+    # preserve CoupledSDEs properties
+    if ds isa CoupledSDEs
```
reykboerner, line 121: @oameye can you check these lines (until line 122) and say whether this is also how you would construct a nonautonomous `CoupledSDEs` from an original `CoupledSDEs` while preserving the noise properties?

The deterministic rule `rss` needs to be passed to the new `CoupledSDEs`, but of course the noise properties should be inherited from the autonomous system

src/r_tipping/RateSystem.jl
```diff
@@ -1,305 +1,389 @@
...
+            else
+                p_new = p_old + rss.forcing_scale[pkey] * 
+                    (f(section_end) - f(section_start))
+            end
+        else
+            p_new = p_old
+        end
+        pdummy[pkey] = p_new
+    end
+    set_parameters!(rss, pdummy, rss.pdummy)
```
reykboerner: @Datseris can you check this line please? 

I renamed the `p_modified!` function to `update_parameters!`, which in the end calls this line with `set_parameters!`. But the first argument of `set_parameters!` should be a system, whereas here in the function space we only have `RateSystemSpecs` (`rss`) available. Does this mean `rss` needs to have an additional field storing `ds`?

src/r_tipping/RateSystem.jl
reykboerner: This is now implemented but we still have the open question of how to deal with the fact that `p_modified!` (now called `update_parameters!`) does not contain a system reference `ds` that can be passed to `set_parameters!(ds, ...)`

.github/workflows/Format.yml
```diff
@@ -3,9 +3,6 @@ name: format-check
 on:
   pull_request:
     types:
-      - opened
-      - reopened
-      - synchronize
       - ready_for_review
```
oameye, line 6: Why?

src/r_tipping/RateSystem.jl
oameye: This won't preserve the noise properties. The `hasproperty` checks all evaluate to false.
I would do:
```julia
if ds isa CoupledSDEs
    IIP = SciMLBase.isinplace(ds)
    prob = referrenced_sciml_prob(ds)
    new_prob = SDEProblem{IIP}(
        rss, prob.g, current_state(ds), (t0, prob.tspan[2]), p0;
        noise_rate_prototype = prob.noise_rate_prototype,
        noise = prob.noise,
    )
    system = CoupledSDEs(new_prob, ds.diffeq, ds.noise_type)
elseif ds isa CoupledODEs
    system = CoupledODEs(rss, current_state(ds), p0; t0)
else
    error("A RateSystem can only be constructed from a CoupledODEs or CoupledSDEs.")
end
```

oameye: Happy to commit. Just let me know if that is handy.

src/r_tipping/RateSystem.jl
```diff
@@ -80,16 +80,34 @@ end
 function RateSystem(
     ds::ContinuousTimeDynamicalSystem,
     forcing_profile::Dict;
-    forcing_start_time=nothing,
-    forcing_duration=nothing,
-    forcing_scale=nothing,
+    forcing_start_time::Dict = Dict(),
+    forcing_duration::Dict = Dict(),
+    forcing_scale::Dict = Dict(),
```
Datseris: Replace `::Dict` with `AbstractDict`

Datseris: Also, all of these should be set to their numeric default value (0, 1, 1) and be removed from the second signature below, see commend below in second signature.

src/r_tipping/RateSystem.jl
```diff
@@ -80,16 +80,34 @@ end
...
+    forcing_duration::Dict = Dict(),
+    forcing_scale::Dict = Dict(),
     t0=initial_time(ds),
     )
 
     if isempty(forcing_profile)
         throw(ArgumentError("`forcing_profile` dictionary must contain at least one entry"))
     end
 
+    if isempty(forcing_start_time)
```
Datseris: Instead of checking if it is empty, check if it is real number.

src/r_tipping/RateSystem.jl
```diff
@@ -130,189 +148,60 @@ function RateSystem(
 end
 
 function RateSystem(ds::ContinuousTimeDynamicalSystem, forcing_profile::ForcingProfile,
-    pidx::Int;
+    pkey;
     forcing_start_time::Real = 0.0, 
     forcing_duration::Real = 1.0,
     forcing_scale::Real = 1.0,
```
Datseris: It is good practice to minimize keyword definition duplication. Don't define keywords at all here, but forward them to the main constructore. simply write `kw...` here and then when calling the main rate system constructor expand them again `RateSystem(new_args; kw...)`.

src/r_tipping/RateSystem.jl
```diff
@@ -130,189 +148,60 @@ function RateSystem(
...
     kwargs...)
 
-    return RateSystem(ds, Dict(pidx => forcing_profile);
-        forcing_start_time = Dict(pidx => forcing_start_time),
-        forcing_duration = Dict(pidx => forcing_duration),
-        forcing_scale = Dict(pidx => forcing_scale),
+    return RateSystem(ds, Dict(pkey => forcing_profile);
+        forcing_start_time = Dict(pkey => forcing_start_time),
+        forcing_duration = Dict(pkey => forcing_duration),
+        forcing_scale = Dict(pkey => forcing_scale),
```
Datseris: This won't be necessary, replace with `kw...`. We make the main constructor aknowledge either a real number or a (filled) dictionary.

src/r_tipping/RateSystem.jl
Datseris: The function will work as well, provided that `current_parameters(rss)` is defined and returns something sensible, a parameter container with identical data layout to the original `ds`. Nevertheless, the `rss` should _anyways_ have a field storing the parent system, this way when the user asks for the frozen system we just return the `ds` field. And then, we should also use this here and give it to `set_parameter!` to avoid any confusion.

It is not yet clear to me what `current_parameter(rss)` returns.

src/r_tipping/RateSystem.jl
```diff
@@ -321,33 +210,36 @@ end
 Returns the parameter container of a [`RateSystem`](@ref) at time `t` (in system time
 units).
 """
-parameters(rs::RateSystem, t) = p_modified(rs.forcing,
-    deepcopy(current_parameters(rs.system)), t)
+function parameters(rs::RateSystem, t)
+    update_parameters!(rs.forcing, t)
+    return rs.forcing.pdummy
+end
```
Datseris, line 198: is `pdummy` the _full_ parameter container, or only the parameters that are forced? THis is something we need to clarify and define. `current_parameters` returns _all_ parameters. What do we do here? we return also all parameters, just that those that are forced have their value at time `t`?

Datseris (commented):
very good progress, let's wrap this up by ensuring all tests pass, as well as writing new tests for the multi parameter case and the SDE case.

.github/workflows/Format.yml
reykboerner: Wanted to mute the format checks while working on it. I saw that applying `format` creates a lot of extra line breaks after kwargs etc. - is there any way to control that?

src/r_tipping/RateSystem.jl
reykboerner: I've copied it, thanks!!

src/r_tipping/RateSystem.jl
reykboerner: It's the full parameter container

src/r_tipping/RateSystem.jl
```diff
@@ -74,70 +72,56 @@ struct RateSystem{S,R} <: ContinuousTimeDynamicalSystem
...
+            forcing_kw[j] = Dict(k => forcing_kw[j] for (k, _) in forcing_profile)
+        else
+            @assert keys(forcing_kw[j]) == keys(forcing_profile) |
+            "Dictionary keys of forcing_$(j) and forcing_profile must be identical."
         end
     end
 
     p0 = deepcopy(current_parameters(ds))
 
     rss = RateSystemSpecs{R,K,T,P}(
```
raphael-roemer: Shouldn't it be `{S,K,T,P}` here?

.github/workflows/Format.yml
oameye: yeah thats happens to keep it the code compact horizontally.
Runic doesn't allow to change rules: https://github.com/fredrikekre/Runic.jl. We changed recently as JuliaFormatter is on life support.

src/r_tipping/RateSystem.jl
raphael-roemer, line 81: Does this really work? Shouldn't we have something like `RateSystemSpecs{S,K,T,P}` ?

oameye: no this will be non-concrete. All structs should be concrete.

Datseris:
Please tag me here explicitly when this is ready to review! I'm turning off notifications until then as I always check and I always realize I shouldn't review yet!

Datseris:
Hello! What is the status here?

oameye:
I am also open the jump in here and push this forward. But don't want to overstep. So just tell me what is appropriate.

Datseris:
@raphael-roemer  have you tried running the example? Because 
```
ERROR: UndefVarError: `parameter` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ c:\Users\gd419\.julia\dev\CriticalTransitions\examples\RateSystem.jl:67
```

raphael-roemer:
I ran a few examples but forgot to export `parameter`, now it should hopefully work fine

Datseris:
`fp2 = ForcingProfile(sin, (-pi, pi))` also errors. 

It's okay, I'll finish this PR and merge tonight because I'd like to present it tomorrow and discuss it with a group of people here!!!

Datseris:
@raphael-roemer seems like the multi parameter case doesn't work/?

The final plot of the Example documentation file you wrote is just:

image

Datseris:
@raphael-roemer there should be for sure a test that tests for the above. The reason this happens is because `parameters` returns always the same internal vector that is modified in place. So `parameters.(Ref(rs2), tvec)` gives always the same entry as it is aliased. `parameters` should operate on a copy of the system. I'll fix this now.

.github/workflows/Format.yml
oameye: @raphael-roemer this need to be reverted!!

examples/RateSystem_old.jl
```diff
@@ -64,7 +64,7 @@ forcing_duration = 105.0 # Time interval over which p(interval) is spread out or
 forcing_scale = 3.0   # Amplitude of the ramping. `p` is then automatically rescaled
 t0 = -70.0            # Initial time of the resulting non-autonomous system (relevant to later compute trajectories)
```
oameye: Should this file be removed?

test/r_tipping/RateSystem.jl
```diff
@@ -18,36 +23,327 @@ function f(u, p, t) # out-of-place
...
+    #             end
+    #             # Minimal smoke: ensure the package loads
+    #             @test true
+    #         catch
+    #             @test true
+    #         end
+    #     else
+    #         @test true
+    #     end
+    # end
```
oameye, line 315: Should this be removed?

Datseris: no, rather enabled back. i'll try to do it after i finish my presentation

.github/workflows/Format.yml
oameye: @Datseris can you also revert this change :)

raphael-roemer:
> `fp2 = ForcingProfile(sin, (-pi, pi))` also errors.
> 
> It's okay, I'll finish this PR and merge tonight because I'd like to present it tomorrow and discuss it with a group of people here!!!

`(-pi, pi)` is `Tuple{Float64, Irrational{:π}`} — -pi becomes Float64 but pi stays Irrational{:π}, so the two tuple elements have different types and don't match `Tuple{T,T}.` So should we change the typing here to allow different types for start and end-time of the interval like this?
```julia
struct ForcingProfile{F,S<:Real,V<:Real}
    profile::F
    interval::Tuple{S,V}
end
```

Datseris:
Hi @oameye , sorry for the late reply. If you want to help you can:

1. ensure the stochastic side examples are runnable and build for the docs
2. add something for the stochastic side in the `tutorial.jl`. That file doesn't run yet due to the rate system problems that i am fixing now. you can comment it out in your attempts to make the docs buildable.

i am presenting tomorrow at 12:45 CET. No worries at all if you don't see this message until then or don't have time to do anything until then.

Tonight i'll do my best to make a workable demo but first there are still some fixes i need to do to the rate system so that it runs as intended so i don't know how far i'll get until tomorrow.

Datseris:
ok, status update: i've finished the rate system and works as intended. I'm merging this PR in now. next steps; I make a new PR that adds my rate phase tipping track return diagrams. Someone else, e.g. Reyk , makes another PR that adds a stochastic system application to the `tutorial.jl`.

reykboerner:
Thanks to all of you for making it work! This is a big step  🚀

Datseris:
(haven't checked if tests pass; I only have capacity to work on `tutorial.jl` at the moment that works)

Datseris:
oh yeah and everyone remember: we need to finish the API.md file.

.github/workflows/Format.yml
oameye: Sad this was merge with this change :(

oameye:
I will clean up the aftermath and work on the rewriting the docs taking into account all new features

Issue #196, opened 2025-08-22
How flexible and compatible is the covariance keyword argument in the CoupledSDEs constructor?


ryandeeley:
I opened this because I wanted to raise a question concerning the utility of the `covariance` keyword argument in the CoupledSDEs constructor, to 1) understand in which cases it should/can be used and to 2) know how compatible it is with the coexisting noise function keyword argument (`g`).

ryandeeley:
This is based off of a short review I made a few months ago of the CoupledSDEs documentation, originally posted in zulip but copied below for all to see: 

Hey, I just read through this: https://juliadynamics.github.io/DynamicalSystemsDocs.jl/dynamicalsystemsbase/stable/CoupledSDEs/#defining-stochastic-dynamics
I think it's great. A quick comment is that when providing a (non-```nothing```) argument to ```covariance```, I think it would add clarity to say what exactly is going on behind the scenes to construct the ```CoupledSDEs```. For instance, for the example discussed within the **Correlated noise** section, is the behind the scenes code either 1) explicitly calling a ```CorrelatedWienerProcess``` possessing this covariance structure and passing this as the noise process to the ```SDEProblem``` when the system is simulated  or 2) constructing the Cholesky decomposition of the covariance matrix to  create the corresponding diffusion matrix  which yields this covariance structure when paired with a regular ```WienerProcess```? I suspect that for the example given it is option 1), because option 2) seems to be then discussed directly below. If it is option 1), I would therefore wonder how restrictive the functionality is here, because ```CorrelatedNoiseProcess``` is not available for every ```NoiseProcess``` that the user might want to pass, unless they personally make a ```CustomNoiseProcess```. I think adding this additional layer of transparency would be helpful, and perhaps it should be made clear that passing a non-```nothing``` ```covariance``` entry [with a ```nothing``` ```g```-function] is really only for specific cases (is it restricted to additive noise, for instance?). The ```SDEProblem``` which arises from prescribing a non-```nothing``` ```g```-function [with a ```nothing``` ```covariance```] seems to be uniquely defined, which is therefore clear.

Another comment is that from reading this, I gather that the user should never pass a non-```nothing``` ```covariance``` entry and a non-```nothing``` ```g``` entry, because the ```SDEProblem``` that would correspond to this is not clear. Is there a warning for this currently, or have I misunderstood it?

[...]

Issue #130, opened 2024-12-24
sgMAM for Multiplicative noise


oameye:
The current implementation only support additive noise. It should be easily extensiable to also support multiplicative noise.

See https://arxiv.org/pdf/[1604.03818](https://arxiv.org/pdf/1604.03818)

PR #320, opened 2026-05-11
feat: implement large deviation for multiplicative noise

### LINKED ISSUE #275: Test gMAM with multiplicative noise
The Freidlin-Wentzell action functional and geometric action functional are well-defined for systems of SDEs with state-dependent, invertible diffusion matrices. I therefore don't understand why this common case is excluded from the `MAM` and `gMAM` functionality. The function `proper_MAM_system` should in particular be modified I suppose, but maybe also the `fw_action` function by the looks of it, to account for the changing diffusion matrix along the inserted path when it happens to be state-dependent.

### LINKED ISSUE #282: `FW_action` returns half the Freidlin–Wentzell action (spurious /2)
## Summary
FW_action on [line 222 of sgMAM.jl](https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/main/src/largedeviations/sgMAM.jl#L222) divides by 2 when it shouldn't, so MLP.action returns half the Freidlin–Wentzell action of the computed path.

## Analysis
The action is computed as:

FW_action(xdot, p) = sum(sum(xdot .* p; dims=1)) / 2
where xdot comes from central_diff!:

xdot[:, 2:(end - 1)] = 0.5 * (x[:, 3:end] - x[:, 1:(end - 2)])
This central difference is not normalized by the grid spacing Δs, so xdot[:,n] ≈ Δs · dϕ/ds. Therefore sum(xdot .* p) is already a valid quadrature of ∫ p · dϕ.

On the zero-energy shell (H = p·f + |p|²/2 = 0), the identity p · dϕ = |p|²/2 · dt holds, so:

∫ p · dϕ  =  ½ ∫ |p|² dt  =  S_FW
The 1/2 from the standard FW definition is already absorbed by this identity. The extra / 2 in FW_action makes the function return S_FW / 2.

This can be cross-checked against the geometric action formula (Eq. 2.4 in [Grafke et al., Multiscale Model. Simul. 17, 2019](https://doi.org/10.1137/18M1211117)):

$$\hat{S}[\phi] = \int_0^L \bigl(|f|,|d\phi| - f \cdot d\phi\bigr),$$

which equals S_FW on the minimizer.

oameye:
resolves #282, #130, #275 

- [ ] `_update_x_general!`: cache the sparse block-tridiagonal pattern. Currently
  `sparse(Iv, Jv, Vv, n, n)` rebuilds the matrix every iteration even though the
  sparsity pattern is fixed. Build `M` and the COO→nzval index map once, then
  overwrite `M.nzval` in place per iteration; reuse a LinearSolve cache so the
  symbolic factorization is shared.
- [ ] `_update_x_diag!`: mirror the additive path's LinearSolve cache. The diagonal
  per-DOF tridiagonal solve currently allocates fresh `Tridiagonal` buffers and
  uses `T_dof \ rhs` without a `LinearProblem` cache; the additive path already
  shows the pattern with `alias_A = true`.
-   [ ] Uniformise the `A` kwarg on `geometric_action(b::Function, ...)` to a single
  type: a callable `x -> a(x)` returning the (uninverted) covariance, with default
  `Returns(LinearAlgebra.I)`. Drop the `A=nothing` and `A::AbstractMatrix` branches;
  constants are wrapped via `Returns(a)`. Mirrors the DSB.jl PR #212 design (single
  covariance entry point) and the discussion in #196.
- [ ] Apply the same treatment to any other public covariance-kwarg surfaces added
  on this branch (e.g. `a_func` in `FreidlinWentzellHamiltonian`, which currently
  dispatches three ways on the return type of `a_func(x)`).

### DIFF
--- a/src/largedeviations/action.jl
+++ b/src/largedeviations/action.jl
@@ -1,30 +1,31 @@
 """
 $(TYPEDSIGNATURES)
 
-Calculates the Freidlin-Wentzell action of a given `path` with time points `time` in a
-drift field specified by the deterministic dynamics `f = dynamic_rule(sys)` and
-(normalized) noise covariance matrix `covariance_matrix(sys)`.
+Calculates the Freidlin-Wentzell action of a given `path` with time points `time` in
+the drift field `f = dynamic_rule(sys)` and (normalized) diffusion tensor of `sys`.
 
-The path must be a `(D x N)` matrix, where `D` is the dimensionality of the system `sys` and
-`N` is the number of path points. The `time` array must have length `N`.
+The path must be a `(D × N)` matrix where `D` is the dimensionality of `sys` and `N`
+is the number of path points. `time` must have length `N`.
 
-Returns a single number, which is the value of the action functional
+Returns a single number, the value of the action functional
 
-``S_T[\\phi_t] = \\frac{1}{2} \\int_0^T || \\dot \\phi_t - f(\\phi_t) ||^2_Q \\text{d}t``
+``S_T[\\phi_t] = \\frac{1}{2} \\int_0^T \\lVert \\dot \\phi_t - f(\\phi_t) \\rVert
+    ^2_{a(\\phi_t)^{-1}} \\, \\mathrm{d}t``
 
-where ``\\phi_t`` denotes the path in state space, ``b`` the drift field, and ``T`` the
-total time of the path. The subscript ``Q`` refers to the
-generalized norm ``||a||_Q^2 := \\langle a, Q^{-1} b \\rangle`` (see `anorm`). Here
-``Q`` is the noise covariance matrix normalized by ``D/L_1(Q)``, with ``L_1`` being the
-L1 matrix norm.
+where ``a(x) = \\sigma(x)\\sigma(x)^\\top`` is the diffusion tensor and the subscript
+``a^{-1}`` denotes the generalized norm
+``\\lVert v\\rVert^2_{a^{-1}} := \\langle v,\\, a^{-1} v\\rangle`` (see [`anorm`](@ref)).
+
+This is the Freidlin-Wentzell **rate function**: it is invariant under uniform rescaling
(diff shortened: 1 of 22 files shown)


src/largedeviations/action.jl
```diff
@@ -157,17 +187,19 @@ end
...
     sqnorm = zeros(size(path, 2))
-    b(x) = drift(sys, x)
     for i in axes(path, 2)
-        # assumes the drift is time independent
-        drift = b(path[:, i])
-        sqnorm[i] = anorm(v[:, i] - drift, A; square = true)
+        xi = view(path, :, i)
+        drift_i = drift(sys, xi)
+        Ai = A isa AbstractMatrix ? A : A(xi)
+        sqnorm[i] = anorm(v[:, i] - drift_i, Ai; square = true)
```
reykboerner, line 202: @oameye is this line the main change that now evaluates the FW action with respect to the state-dependent covariance matrix? Trying to get an overview of this PR

src/largedeviations/minimize_action.jl
```diff
@@ -109,10 +109,3 @@ function fix_ends(x::Matrix, x_i, x_f)
     m[:, end] = x_f
     return m
 end;
-
-"""
-    action_minimizer(sys::ContinuousTimeDynamicalSystem, x_i, x_f, T::Real; kwargs...)
-
-Alias for [`minimize_action`](@ref).]
-"""
-const action_minimizer = minimize_action
```
reykboerner, line 118: Yes, good to remove this

src/largedeviations/action.jl
oameye: Yes. but the PR needs quite still some work. It was coded up by a bot and it did a shitty job. We need to have a uniform choice of covariance matrix on which we settled before. I will ask a review when it is ready

reykboerner: Okay, thanks for fixing this :)

oameye:
superseded by #338

PR #331, opened 2026-05-14
fix(transitions): forward sys.diffeq and reseed per call

### LINKED ISSUE #283: transitions() does not forward sys.diffeq kwargs to solve()
## Bug description

`transitions()` (and `transition()`) do not forward the `sys.diffeq` keyword arguments (e.g. `maxiters`, `dt`, `dtmin`, `force_dtmin`, `abstol`, `reltol`) to the internal `solve()` call. Only the algorithm is extracted via `solver(sys)`.

This means that setting solver options through `diffeq` when constructing a `CoupledSDEs`:

```julia
sde = CoupledSDEs(f!, u0, p; noise_strength=σ, diffeq=(alg=SOSRA(), maxiters=5_000_000))
```

has **no effect** on `transitions(sde, x_i, x_f, N)` — the solver runs with default `maxiters` (100,000).

## Where the bug is

In [`src/trajectories/transition.jl`, lines 152–158](https://github.com/juliadynamics/CriticalTransitions.jl/blob/main/src/trajectories/transition.jl#L152-L158):

```julia
sim = solve(
    ensemble,
    solver(sys),   # only extracts the algorithm
    EnsembleAlg;
    callback=cb_ball,
    trajectories=min(N, Nmax),
    kwargs...,     # these are the caller's kwargs, NOT sys.diffeq
)
```

And `solver(sys)` in `src/sde_utils.jl` line 53:
```julia
solver(ds::ContinuousTimeDynamicalSystem) = ds.integ.alg
```

The same issue exists in `transition()` (line 42).

## Impact

For stiff systems where `Δ/κ >> 1` (e.g. underdamped KPO with `Δ/κ = 50`), the default `maxiters` is far too low. Trajectories silently fail (`retcode != Terminated`) and transition rates are biased low. In my case, the default gave ~56/200 successful transitions while forwarding `maxiters=5_000_000` correctly gave ~155/200.

## Minimal reproducer

```julia
using CriticalTransitions

function stiff_drift!(du, u, p, t)
    Δ, K, G, κ = p
    r² = u[1]^2 + u[2]^2
    du[1] = -κ/2 * u[1] + (-Δ + G + 2K * r²) * u[2]
    du[2] = -κ/2 * u[2] + (Δ + G - 2K * r²) * u[1]
    return nothing
end

p = [0.01, 0.001, 0.015, 0.0002]  # Δ/κ = 50
u0 = [3.5355, 0.0118]

# maxiters is set in diffeq but has no effect on transitions()
sde = CoupledSDEs(stiff_drift!, u0, p;
    noise_strength=sqrt(0.0002 * 8.5),
    diffeq=(alg=SOSRA(), maxiters=5_000_000),
)

ens = transitions(sde, u0, -u0, 50;
    radii=(0.2, 0.2), tmax=5e7, Nmax=50, show_progress=false,
)
# Many trajectories fail because solve() uses default maxiters=100_000
```

## Suggested fix

Extract `sys.diffeq` kwargs and merge them into the `solve()` call, letting caller kwargs take precedence:

```julia
# In transitions():
diffeq_kw = Dict{Symbol,Any}(k => v for (k, v) in pairs(sys.diffeq) if k !== :alg)
merged = merge(diffeq_kw, Dict{Symbol,Any}(kwargs...))

sim = solve(
    ensemble,
    solver(sys),
    EnsembleAlg;
    callback=cb_ball,
    trajectories=min(N, Nmax),
    merged...,
)
```

The same pattern should be applied to `transition()` at line 42.

### LINKED ISSUE #284: `transitions` uses same random seed when parallelized
**Describe the bug**
A while ago I noticed when running `transitions` with several threads that the trajectories are partly identical

I assume the same random seed is chosen for multiple threads

oameye:
- #283: forward sys.diffeq kwargs (maxiters, abstol, reltol, ...) into solve() in both transition() and transitions(); caller kwargs take precedence via diffeq_kw splatted before kwargs.
- #284: derive a fresh master seed via rand(UInt64) each call to transitions(), then a unique per-sim seed from master_seed + ctx.sim_id + ctx.repeat. Parallel sims no longer share RNG state and independent calls produce independent ensembles. sys's stored seed is no longer consulted in transitions().
- Add seed::Union{Nothing,Integer}=nothing kwarg on both functions for opt-in reproducibility, mirroring DSBase reinit!(; seed=...).

Closes #283
Closes #284


src/trajectories/transition.jl
```diff
@@ -32,14 +34,19 @@ function transition(
...
     prob, cb_ball = prepare_transition_problem(
         sys, (x_i, x_f), radii, radius_directions, tmax
     )
+    if seed !== nothing
+        prob = remake(prob; seed = UInt64(seed))
+    end
 
-    sim = solve(prob, solver(sys); callback = cb_ball, kwargs...)
+    diffeq_kw = NamedTuple{filter(!=(:alg), keys(sys.diffeq))}(sys.diffeq)
+    sim = solve(prob, solver(sys); callback = cb_ball, diffeq_kw..., kwargs...)
```
Datseris, line 49: Why are we using `solve` instead of `trajectory` or `step!`?

oameye: `trajectory` explicitly doesn't support callbacks. From the [DSBase docstring][trajectory] (continuous case): *"it doesn't play well with callbacks, use `DifferentialEquations.solve` if you want a trajectory that works with callbacks."* The whole point of `transition()` is to terminate on a condition (entering the ball around `x_f`), which the [`DiscreteCallback` in `prepare_transition_problem`][cb] does for us. Without it, we'd have to write a manual `step!` loop polling `subnorm(current_state(sys) - x_f)` after every step. Workable, but it re-implements behaviour `solve` gives us for free.

Also `transitions()` is built on [`EnsembleProblem`][ensemble], which gives us composable `EnsembleAlg` parallelism (`EnsembleThreads`, `EnsembleDistributed`, `EnsembleSerial`) plus `prob_func`/`output_func` for per-trajectory reseeding and early filtering. There's no equivalent in the DS.jl API; [`ParallelDynamicalSystem`][pds] evolves multiple states of the same system at the *same times*, not independent trajectories. Replacing it would mean rolling our own threaded loop.

A `step!`-based rewrite is technically possible but would essentially re-implement a subset of `EnsembleProblem` + callbacks in our own code.

[trajectory]: https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/v3.16.0/src/core/trajectory.jl#L34-L38
[cb]: https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/135ca0ab96ac197ed43f977f0e4da39351bc7eea/src/trajectories/transition.jl#L62-L64
[retcode]: https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/135ca0ab96ac197ed43f977f0e4da39351bc7eea/src/trajectories/transition.jl#L50
[ensemble]: https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/135ca0ab96ac197ed43f977f0e4da39351bc7eea/src/trajectories/transition.jl#L165-L174
[pds]: https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/v3.16.0/src/derived_systems/parallel_systems.jl#L9-L25

Datseris: > The whole point of `transition()` is to terminate on a condition (entering the ball around `x_f`),

This is what we do in ChaosTools.jl for return time statistics: https://github.com/JuliaDynamics/ChaosTools.jl/blob/main/src/rareevents/mean_return_times/mrt_continuous.jl#L32-L88 . It is worth checking if this is substantially faster or not than the callback. Early on in DS.jl I've moved away from callbacks for this kind of stuff, also for the Poincare, because I found the noncallback code to be much faster. I haven't checked in a while if this is still true.

This issue suggested also formalizing in terms of some sort of iterator or derivative dynamical system: https://github.com/JuliaDynamics/ChaosTools.jl/issues/319 

But I understand why this would be a lot of effort not currently necessary. Regarding the interfacing with EnsembleProblem, I guess that's another benefit.

Datseris: But all good, it was just a comment.

PR #352, opened 2026-06-03
Add reverse rate forcing; add rate phase diagram

Datseris:
Ongoing work. I am too tired to finish now!


Datseris:
Okay, I have added the rate tipping diagram function. BUT. it doesn't work as intended. Not clear yet where the bug is or what I am doing wrong. Final plot looks like:

image


but it should look like:

image


as you can see it doesn't detect at all the always tip state.

Datseris:
any help appreciated.

Datseris:
not sure why there are so many conflicts. Just focus on the file : `r_tipping_phase_diagrams.jl`

oameye:
> not sure why there are so many conflicts.

it is because in #356 I formatted and made the tests pass. I will merge main.

oameye:
image

Datseris:
@oameye thanks a lot for the fix! but for me to understand it you need to write here what was the original problem and what you did to fix it. I see some new deepcopies in the code, but we already deep copy so much in the rate system, so its not clear to me if not deep copying is necessary

oameye:
> @oameye thanks a lot for the fix! but for me to understand it you need to write here what was the original problem and what you did to fix it. I see some new deepcopies in the code, but we already deep copy so much in the rate system, so its not clear to me if not deep copying is necessary


Tbh I was just makes following hunges until it worked. The problem was that the return phase of `rate_track_return_tip` never reversed the forcing. After the forcing ramp ended, the parameter stayed at its displaced value instead of going back, so the system always relaxed onto the shifted system's attractor. That made `return_id ≠ start_id` almost everywhere, and the diagram came out as nearly all "tip".

The fix was to reverse the forcing for the return phase `set_forcing_scale!(rs, -dp)` so the parameter is actually driven back to its starting value.

also at [2.0, 1, 0.3] the stommel only had one attractor.

the relax phase and deepcopy where just to see if it solved the problem. I reverted it.

Datseris:
> The fix was to reverse the forcing for the return phase set_forcing_scale!(rs, -dp) so the parameter is actually driven back to its starting value.

Okay I think I get the problem; I've implemented the reverse forcing at the level of the `RateSystem`. I just forgot to turn it on at the start of the function. We shouldn't do what you do there, adding inverse forcing scale. I'll fix it later.

Datseris:
> also at [2.0, 1, 0.3] the stommel only had one attractor.

that was also intended. i'll revert this as well.

oameye:
> We shouldn't do what you do there, adding inverse forcing scale. 

Ha okay lol then I misunderstood the intended starting point

Datseris:
ummm what the hell happened in this pr?!?! in this commit: https://github.com/JuliaDynamics/CriticalTransitions.jl/pull/352/changes/12cbbca00abafa1a88a003b85af4b59eeb1e06f7 I've added the `reverse` parameter to the rate system. In the current stage of the source code this doesn't exist?!? @oameye

Datseris:
it may well be a problem on my end. i'm struggling the hell out of my life to use git in VSCode. I was used to gitkraken. I'm switching there now.

oameye:
sorry maybe my merged with main broke things

PR #355, opened 2026-06-04
docs: rewrite the manual

oameye:


reykboerner:
Good idea to add the developer notes page and explain the noise strength convention. Are you planning to rewrite other things in this PR @oameye?

oameye:
@reykboerner I propose that I push this until the docs pass without `warn_only=true`. So that we have some quality gates in place. Afterwards, we can make a couple of smaller PR to get the docs in a good place. It would be nice to have faster pace on the reviews to get the docs to a proper state.

reykboerner:
> @reykboerner I propose that I push this until the docs pass without `warn_only=true`. So that we have some quality gates in place. Afterwards, we can make a couple of smaller PR to get the docs in a good place. It would be nice to have faster pace on the reviews to get the docs to a proper state.

Sure, that's fine!

PR #361, opened 2026-06-08
docs: update Maierstein example

### LINKED ISSUE #357: docs: quasipotential value
If the QP at the saddle is expected to be 1/3, isn't the computed value of 0.48 a bit far off?

https://github.com/JuliaDynamics/CriticalTransitions.jl/blob/6761c18aa3813ac09083e578badc661fe716b34f/examples/quasipotential_maierstein.jl#L27

oameye:
resolve  #357

remove false claim of the action being 1/3. Instead, validate against the analytical curve
$$
U_A(u, 0) = 2\,\bigl(V(u) - V(-1)\bigr) = \tfrac{1}{2}(1 - u^2)^2
$$
 from [Maier and Stein 1993](https://doi.org/10.1103/PhysRevE.48.931). Also compare against gmam.


reykboerner:
Great @oameye :)

PR #386, opened 2026-09-18
Add tipping probability calculation

### LINKED ISSUE #374: Rudimentary tipping probabilities code
I've removed the tipping probabilities function from Attractors.jl in v2 so that it can be moved here. I am pasting here the raw docs and code and later I'll do a proper PR:

```

## Tipping points
This page discusses functionality related with tipping points in dynamical systems with known rule. If instead you are interested in identifying tipping points in measured timeseries, have a look at [TransitionIndicators.jl](https://github.com/JuliaDynamics/TransitionIndicators.jl).

@docs
tipping_probabilities


```

```julia
export tipping_probabilities

"""
    tipping_probabilities(BoA_before::ArrayBasinsOfAttraction, BoA_after::ArrayBasinsOfAttraction) → P
    tipping_probabilities(basins_before, basins_after) → P

Return the tipping probabilities of the computed basins before and after a change
in the system parameters (or time forcing), according to the definition of [Kaszas2019](@cite).

The input `basins` are integer-valued arrays, where the integers enumerate the attractor,
e.g. the output of [`basins_of_attraction`](@ref).

The second function signature exists for backwards compatibility. 

## Description
Let ``\\mathcal{B}_i(p)`` denote the basin of attraction of attractor ``A_i`` at
parameter(s) ``p``. Kaszás et al [Kaszas2019](@cite) define the tipping probability
from ``A_i`` to ``A_j``, given a parameter change in the system of ``p_- \\to p_+``, as

#```math
P(A_i \\to A_j | p_- \\to p_+) =
\\frac{|\\mathcal{B}_j(p_+) \\cap \\mathcal{B}_i(p_-)|}{|\\mathcal{B}_i(p_-)|}
#```
where ``|\\cdot|`` is simply the volume of the enclosed set.
The value of `` P(A_i \\to A_j | p_- \\to p_+)`` is `P[i, j]`.
The equation describes something quite simple:
what is the overlap of the basin of attraction of ``A_i`` at ``p_-`` with that of the
attractor ``A_j`` at ``p_+``.
If `basins_before, basins_after` contain values of `-1`, corresponding to trajectories
that diverge, this is considered as the last attractor of the system in `P`.
"""
function tipping_probabilities(basins_before::AbstractArray, basins_after::AbstractArray)
    @assert size(basins_before) == size(basins_after)

    bid, aid = unique.((basins_before, basins_after))
    P = zeros(length(bid), length(aid))
    N = length(basins_before)
    # Make -1 last entry in bid, aid, if it exists
    put_minus_1_at_end!(bid); put_minus_1_at_end!(aid)

    # Notice: the following loops could be optimized with smarter boolean operations,
    # however they are so fast that everything should be done within milliseconds even
    # on a potato
    for (i, ι) in enumerate(bid)
        B_i = findall(isequal(ι), basins_before)
        μ_B_i = length(B_i) # μ = measure
        for (j, ξ) in enumerate(aid)
            B_j = findall(isequal(ξ), basins_after)
            μ_overlap = length(B_i ∩ B_j)
            P[i, j] = μ_overlap / μ_B_i
        end
    end
    return P
end

function put_minus_1_at_end!(bid)
    return if -1 ∈ bid
        sort!(bid)
        popfirst!(bid)
        push!(bid, -1)
    end
end


tipping_probabilities(
    BoA_before::ArrayBasinsOfAttraction,
    BoA_after::ArrayBasinsOfAttraction
) = tipping_probabilities(BoA_before.basins, BoA_after.basins)

```

oameye:
Closes #374.

Ports the basin-overlap `tipping_probabilities` functionality from Attractors.jl into CriticalTransitions.jl.

- supports raw integer basin arrays and `ArrayBasinsOfAttraction`
- preserves the historical probability-matrix ordering, with divergent label `-1` placed last
- computes the overlap counts in a single pass rather than repeated `findall`/set intersections
- adds targeted tests, documentation, and the Kaszás et al. (2019) reference


Datseris (approved):

src/r_tipping/tipping_probabilities.jl
```diff
@@ -0,0 +1,85 @@
...
+"""
+    tipping_probabilities(rs::RateSystem, grid)
+
+Return the basin-overlap tipping probabilities between the past and future
+limit systems of the RateSystem `rs`, following [Kaszas2019](@cite).
+
+The `grid` argument specifies the state space grid over which the basins of
+attraction are computed.
+"""
+function tipping_probabilities(rs::RateSystem, grid)
```
reykboerner: I think this high-level function would be nice, what do you think? Just needs to be finished where it's commented

oameye: `RateSystem + grid` does not uniquely specify how to compute the basins, since Attractors.jl still needs a `BasinMap`.

reykboerner: Yes, true. We could make this function only work with `BasinMapRecurrences` and refer to the more low-level method for other uses. Or maybe this function is too high-level because it might also need to perform continuation to track the attractors.

oameye: well yeah I think often one needs the lower level API anyway. So maybe we are just adding complexity with this high-level API

reykboerner (approved):

Issue #339, opened 2026-05-23
Performance audit: large deviations module


oameye:
Made with an LLM

Sampling profile + BenchmarkTools per-call measurements across every entry point in `src/largedeviations/`. Two systemic issues account for nearly all per-iteration cost in the non-IIP paths; one outright bug surfaced during setup. Details below.

Tools used: BenchmarkTools.jl `@benchmark` (median time / total allocs / bytes), `Base.Profile` sampling profile, `InteractiveUtils.code_warntype`. Path size `N=100`, dim `Nx=2` unless noted.

## Systems exercised

| Tag | Construction | Noise class | sgMAM cache | gMAM `assemble_rhs!` |
|---|---|---|---|---|
| **KPO (IIP)** | `FreidlinWentzellHamiltonian{true,2}(H_x!, H_p!)` (user-written, `const` globals) | constant identity-style | Decoupled | n/a |
| **MS-add (OOP auto)** | `FreidlinWentzellHamiltonian(CoupledSDEs)` / additive identity | constant diagonal | Decoupled | `Val{true}` branch |
| **MS-mdiag** | additive=false, `σ(u)=[0.3+0.1u₁², 0.3+0.1u₂²]` | state-dep diagonal | Decoupled | `Val{false}` branch (FD ∂a) |
| **MS-mcoup** | additive=false, `σ(u)=[1 0.3u₁; 0.2u₂ 1]` | state-dep coupled | **Sparse-coupled** | `Val{false}` branch |

## Top-level wall-clock (median)

| Entry point | Setup | iters | Time | MiB | Allocs |
|---|---|---:|---:|---:|---:|
| `minimize_geometric_action(sys::FWHam, …, GeometricGradient)` | KPO IIP | 500 | 4.5 ms | 6.76 | 18 809 |
| same | KPO IIP | 200 | 1.78 ms | 2.73 | 7 709 |
| same | MS-add OOP auto | 200 | 9.38 ms | 16.82 | 308 916 |
| same | MS-mdiag OOP auto | 200 | 23.5 ms | 54.42 | 1 310 541 |
| same | MS-mcoup OOP auto (sparse cache) | 200 | 50.7 ms | 95.88 | 1 351 187 |
| `minimize_geometric_action(sys::FWHam, …, AdaptiveGeometricGradient(probe=50))` | KPO IIP | 400 | 7.87 ms | 10.74 | 26 071 |
[...]

PR #390, opened 2026-09-18
Cache Jacobian in Onsager-Machlup action

oameye:
Addresses the `om_action` performance item in #339.

The Onsager–Machlup correction previously called `div_drift(sys, x)` at both ends of every trapezoid segment. Each call rebuilt `jacobian(sys)`, and every interior path point was evaluated twice.

This change:
- constructs `jacobian(sys)` once per `om_action` call;
- reuses the system parameters directly;
- carries the previous divergence through the trapezoid loop, so each path node is evaluated exactly once;
- leaves the mathematical trapezoid rule and public API unchanged.

Performance certification with #395/#407:
- `om_action`: roughly **1.03 ms / 20,532 allocations → 121 μs / 3,665 allocations** on the isolated certification stack;
- final all-slices #427: **149.047 μs / 139,912 B / 3,665 allocations**.

Clean head `e37d2c3749a46b27aad205842d53d6b6908fc217` is fully green across Tests latest+LTS, Documentation, JET, Format, Spell, and Benchmark Tracking.

Refs #339, #395, #427.


oameye:
Certification result from measurement PR #407 is decisively positive. On the tracked Onsager–Machlup action benchmark, caching `jacobian(sys)` and carrying the previous divergence reduces the kernel from ~1.031 ms / 1,362,408 B / 20,532 allocations on the #395 baseline to 121.253 μs / 139,912 B / 3,665 allocations: about 8.5× faster, ~89.7% less allocated memory, and ~82.2% fewer allocations. #407 is fully green across Tests, Documentation, JET, Format, Spell, and Benchmark Tracking. Retain #390; the acceptance criterion is clearly met.

Refs #339.

PR #391, opened 2026-09-18
Fuse decoupled sgMAM momentum update

oameye:
Superseded by #428, which contains the exact #391 momentum-fusion change together with the overlapping #393 Thomas-solver change as one clean conflict-free commit on frozen `main`.

The #391 endpoint-`p` semantics were audited and preserved in #428: endpoint `lambda` is zeroed before constructing endpoint momentum. #428 is fully green across Tests latest+LTS, Documentation, JET, Format, Spell, and Benchmark Tracking.

Closing without merge so the production landing has one reviewable PR instead of a known `sgmam_kernels.jl` conflict.

Refs #339, #393, #428.


oameye:
Isolated certification from measurement PR #408 is fully green across Tests, Documentation, JET, Format, Spell, and Benchmark Tracking. The allocation signal is positive and clean on the multiplicative diagonal sgMAM workload: 73,085 allocations / 2,928,192 B here versus 89,081 allocations / 3,838,736 B on the contemporaneous #395+#390 control (#407), i.e. ~18.0% fewer allocations and ~23.7% less allocated memory. Wall-clock values vary materially across the hosted runners, so I would not claim a timing speedup from this isolated comparison.

The earlier endpoint-`p` semantic concern is also resolved: the baseline zeros endpoint λ before forming p, and the fused loop preserves that ordering pointwise. Retain #391 as a modest, semantically equivalent allocation/memory improvement; the much larger sgMAM gains come from the cumulative #393/#394 stack.

Refs #339.

PR #392, opened 2026-09-18
Support StateSpaceSet input in gMAM

### LINKED ISSUE #264: make `gmam` work with `StateSpaceSet`

oameye:
Closes #264.

`minimize_geometric_action` already accepts state-space-set paths on the Hamiltonian sgMAM route, while the direct `CoupledSDEs` gMAM route still required a `D × N` matrix.

This PR now treats that as an API-boundary canonicalization rather than carrying `StateSpaceSet` through the numerical kernels:
- direct gMAM accepts any `StateSpaceSets.AbstractStateSpaceSet`;
- a shared `_path_matrix` helper converts the point-major state-space-set representation directly into the mutable dense `D × N` representation used internally, with one final matrix allocation and no intermediate `Matrix(Matrix(path)')` conversion;
- the existing sgMAM `StateSpaceSet` overload is generalized to `AbstractStateSpaceSet` and uses the same canonicalizer;
- the Maier–Stein regression covers both a concrete `StateSpaceSet` and a `SubStateSpaceSet` view, checks exact orientation, dispatch, and endpoint preservation.

The gMAM/sgMAM numerical algorithms remain matrix-based and are otherwise unchanged. The branch was rebuilt as one commit on current `main`, preserving the already-landed #398 numerical-diagonal fast path.

Clean head: `73ce678535bb8d4ade2c4cc45a73a73719c95ddc`.

No merge without explicit approval.


Datseris:
why do you need to make everything a matrix to begin with, instead of a `AbstractVector{<:AbstractVector}`, which a `StateSpaceSet` already subtypes.

oameye:
I vaquily remember when I was deciding this that `StateSpaceSet` was quite a lot slower in doing the operations (#148).  I rather had that the instanton was presented by a `StateSpaceSet`, but likely this will require a larger rewrite if we don't want to loose the performance

PR #396, opened 2026-09-18
Reuse coupled sgMAM symbolic LU factorization

oameye:
Addresses the remaining sparse-coupled linear-solve hotspot from #339.

The coupled sgMAM block matrix has a fixed CSC sparsity pattern for the lifetime of the cache; only `M.nzval` changes between iterations. The existing path routes every solve through `LinearSolve.reinit!` + `solve!`, which rebuilds the sparse factorization machinery on each update.

This PR:
- initializes one UMFPACK LU factorization when the coupled cache is built;
- keeps the matrix sparsity pattern unchanged;
- updates only numeric values in `M.nzval` as before;
- refreshes the numeric LU with `lu!(F, M)`, reusing UMFPACK's symbolic factorization;
- solves into a caller-owned reusable vector.

The discretization, matrix assembly, state-dependent diffusion handling, and decoupled path are unchanged. Existing multiplicative-noise end-to-end tests exercise the coupled path; Benchmark Tracking provides the performance check.

Refs #339.


oameye:
Benchmark result: the direct UMFPACK reuse path does not produce a material improvement on the tracked coupled workload. `Simple geometric minimal action/multiplicative off-diagonal` is 7.80 ms / 97,094 allocations here versus roughly 8.0 ms / 96,965 allocations on recent `main`. The timing delta is only a few percent and allocations are slightly higher, so this does not justify bypassing LinearSolve.

This strongly suggests the existing LinearSolve cache is already effectively reusing the sparse symbolic setup. Closing this as a useful negative experiment; the next optimization should target the remaining state-dependent diffusion / per-point dense work instead.

Refs #339.

PR #397, opened 2026-09-19
Reuse diffusion buffers in auto Hamiltonians

oameye:
Experimental follow-up to #394 for the remaining state-dependent diffusion allocations from #339.

The branch contains #394 plus one incremental diffusion-buffer commit; it targets `main` only so the full Tests/JET/Docs/Benchmark workflows run. The incremental effect is measured against #394's recorded benchmark run.

The ordinary `a(x)` contract is preserved exactly. For state-dependent auto-derived Hamiltonians, an internal typed wrapper additionally exposes the underlying diffusion function and normalization scale so caller-owned matrices can be filled directly.

This draft currently:
- reuses one diffusion matrix across path columns in auto `H_p`;
- reuses two diffusion matrices across the central finite-difference evaluations in auto `H_x`;
- leaves constant-noise paths unchanged;
- keeps `sys.a(x)` delegating to the existing trace-normalized callable, preserving diagonal/coupled classification and compatibility.

The purpose of this draft is to certify whether removing repeated `(σ σᵀ)/s` materialization gives a measurable post-#394 gain before broadening the same mechanism to direct gMAM.

Refs #339.


oameye:
Negative benchmark result: this buffering approach should not be kept. Relative to #394, multiplicative sgMAM regressed from 1.418 ms / 2,069 allocs to 2.202 ms / 2,952 allocs for diagonal noise, and from 6.433 ms / 35,288 allocs to 6.629 ms / 35,666 allocs for off-diagonal noise. The dense caller-owned diffusion-buffer abstraction is therefore counterproductive for these small/static tensors. Closing as an informative experiment; no production change should be carried forward.

PR #398, opened 2026-09-19
Use diagonal fast path in direct gMAM

oameye:
Performance follow-up for #339 in direct gMAM.

State-dependent matrix-valued diagonal diffusion is validated as diagonal during setup, but `_lambda_theta!` previously recognized only `LinearAlgebra.Diagonal`. Static/dense diagonal matrices therefore fell through to a fresh dense LU at every interior path point and iteration.

This change keeps the existing pointwise diffusion representation and extends the scalar solve path to numerically diagonal matrices via `_isdiag_numerical(a_i)`, using `a_i[k,k]` directly. The check is pointwise, so it does not rely on a potentially stale reference-state classification. General off-diagonal diffusion and the cached constant non-diagonal LU path are unchanged.

Benchmark certification on the tracked multiplicative-diagonal direct-gMAM workload:
- allocations: 63,623 -> 42,095 (-33.8%);
- allocated bytes: 1,927,568 -> 1,123,856 (-41.7%).

The off-diagonal control remains exactly 95,694 allocations, confirming that the change only affects the diagonal path. Cross-run hosted-runner wall time is noisy, so the performance claim is intentionally limited to the substantial allocation/memory reduction.

A focused regression checks that a dense numerically diagonal covariance produces the same `lambda`, `theta`, and solved vectors as the canonical `Diagonal` representation.

The clean one-commit head `957229af2ff6b7f235edc98c5bfdb24a296dc2af` is fully green across Tests, Documentation, JET, format, spell, and Benchmark Tracking.

Refs #339.


oameye (commented):
Benchmark certification is positive on the allocation axis. Relative to the current main/#395 baseline for the same tracked workload, `Geometric minimal action/multiplicative diagonal` drops from 63,623 allocations / 1,927,568 B to 42,095 allocations / 1,123,856 B: -33.8% allocations and -41.7% allocated bytes. The off-diagonal control remains exactly 95,694 allocations, as expected. Wall-clock differs across hosted runners, so I would not use the cross-run timing delta as evidence here; the deterministic allocation reduction is substantial enough to retain this fast path pending the full test matrix.

PR #414, opened 2026-09-19
Keep static coupled sgMAM metric solves static

oameye:
Production performance slice for #339, one clean commit on current `main`.

This PR targets only the per-path-point covariance solves in coupled sgMAM `update_p!`:
- when state-dependent `a(x_t)` is a `StaticMatrix`, preserve that structure and use a small static inverse factor instead of materializing/factorizing a heap `Matrix`;
- solve the three momentum RHS vectors through a static kernel;
- still copy `a(x_t)` into the existing `a_at[t]` buffer because the coupled `update_x!` path consumes those covariance blocks;
- dynamic matrix-valued covariance retains the dense-LU fallback;
- constant covariance retains the precomputed-LU path.

Focused regressions certify the static solve against the dense reference and verify that dynamic matrices keep the LU fallback.

Isolated fixed 10-iteration multiplicative sgMAM certification:
- coupled baseline: ~1.152 ms / 1.32 MB / 4,884 allocations;
- #414: ~904 μs / 1.116 MB / 924 allocations;
- diagonal control unchanged (~175 μs / 416 allocations).

That is about 21.5% faster with ~81% fewer allocations for the coupled case. The clean #414→#415 production tree was fully green across Tests latest+LTS, Documentation, JET, Format, Spell and Benchmark Tracking before landing began, and the final #427 integrated stack reaches 499.5 μs / 88,744 B / 679 allocations for the same fixed coupled workload.

Rebased cleanly after #389/#390/#394/#395/#398/#399/#400 landed. The conflict was only include-list overlap; all landed changes are preserved.

Clean head: `9e042720dd00d0e069c4fa0d76cfdf5022a5c980`, exactly one commit on current `main`.

No public API changes. No merge without explicit approval.

Refs #339, #413, #415, #427.


oameye:
Benchmark result against the fixed #413 cumulative sgMAM baseline:

- coupled multiplicative sgMAM, fixed 10 iterations: ~1.152 ms / 1.32 MB / 4,884 allocs → **904 μs / 1.116 MB / 924 allocs**
- diagonal control remains essentially unchanged at ~175 μs / 30.6 KB / 416 allocs

So preserving static covariance solves in coupled `update_p!` gives about **21.5% lower time and 81% fewer allocations** on the target coupled workload without moving the diagonal control. Full Tests, Docs, JET, Format, Spell, and Benchmark Tracking are green. Retaining this optimization.

PR #417, opened 2026-09-19
Certify combined coupled sgMAM fast paths

oameye:
Cumulative certification stack for #339.

This branch combines the two independently positive coupled-sgMAM experiments:

- #414: keep state-dependent static covariance solves static inside coupled `update_p!`;
- #415: replace the global sparse `update_x!` solve by the structurally exact block-Thomas/block-Cholesky recurrence when applicable, retaining the existing sparse solver as fallback.

The branch inherits the fixed 10-iteration multiplicative diagonal/off-diagonal benchmark cases from the #413 certification stack and includes both focused regression suites. It also updates the two pre-existing concrete cache-type assertions to the new block wrapper; no user-facing API is changed.

Acceptance criterion: the coupled fixed benchmark must improve materially over both isolated baselines without regressing the diagonal control, and full Tests/Docs/JET/Format/Spell must remain green.

Reference measurements:
- #413 cumulative baseline: ~1.152 ms / 1.32 MB / 4,884 allocs coupled;
- #414 static metric solve: ~904 μs / 1.116 MB / 924 allocs coupled;
- #415 block Thomas alone: ~854 μs / 289 KB / 4,639 allocs coupled.

Measurement/certification only; do not merge this PR directly.

PR #418, opened 2026-09-19
Certify unscaled block-Thomas coupled sgMAM

oameye:
Follow-up certification on top of the #417 cumulative coupled-sgMAM stack.

This changes only the block-Thomas recurrence and its dense-reference regression. Instead of row-scaling each block equation by `1 / lambda_i^2`, it applies block Thomas directly to the original sparse-system blocks:

`S_i = D_i - eps^2 * q_i * q_{i-1} * inv(S_{i-1})`, with `q_i = lambda_i^2`.

This is algebraically equivalent for nonzero `q`, but is numerically preferable because it:
- avoids amplifying small `lambda_i` through division by `q_i`;
- naturally handles zero interior `lambda_i` without falling back;
- preserves symmetric positive-definite Schur blocks whenever the physical covariance blocks are SPD;
- leaves the original sparse solver as fallback for non-symmetric metrics or failed Cholesky.

The updated dense-reference test deliberately includes a zero interior lambda and compares against the original nonsymmetric block-tridiagonal system directly.

This is measurement/certification only. It should replace the row-scaled block recurrence only if full CI is green and the fixed coupled sgMAM benchmark is neutral or better relative to #417.


oameye:
Benchmark result: retain the unscaled recurrence.

Fixed 10-iteration multiplicative sgMAM:
- off-diagonal coupled: **645.401 μs / 88,744 B / 679 allocations**;
- diagonal control: **180.369 μs / 30,624 B / 416 allocations**.

The coupled result is performance-neutral/slightly faster than #417's row-scaled 654.467 μs / 88,744 B / 679 allocations, with identical memory/allocation counts. The diagonal difference is hosted-runner noise; its memory/allocation counts are identical and this PR does not touch the decoupled path.

Therefore the unscaled recurrence is strictly preferable architecturally: it solves the original lambda^2-weighted block system directly, avoids `1/lambda_i^2`, and the dense-reference regression covers zero interior lambda. JET/Format/Spell and Benchmark Tracking are green; LTS Tests are green, with Julia-latest Tests/Docs still finishing at this checkpoint.

oameye:
Certification complete. This unscaled block-Thomas formulation is fully green across Tests (latest + LTS), Documentation, JET, Format, Spell, and Benchmark Tracking. Its production delta has been transferred into clean one-commit stacked PR #415 on top of clean #414. Closing this cumulative measurement branch without merge.

PR #419, opened 2026-09-19
Reuse block Cholesky wrappers in coupled sgMAM

oameye:
Follow-up certification for #339, stacked on the numerically safer unscaled block-Thomas branch (#418).

The block cache already stores one `Cholesky` wrapper per reusable Schur matrix, and each wrapper points to that same matrix. During every block-Thomas sweep, `cholesky!` mutates the Schur matrix in place, but the current implementation then stores the newly returned wrapper back into `factors[i]`.

That replacement is unnecessary: the existing wrapper still references the freshly factorized matrix and has the same `:L` orientation. This change therefore removes only `factors[i] = C`; the returned factor is still checked with `issuccess(C)` before any solve proceeds.

For the fixed coupled benchmark there are 58 interior blocks × 10 iterations = 580 factorization sites, while #417 reports 679 total allocations. This is therefore a targeted test of the dominant remaining allocation signature.

No arithmetic, fallback behavior, public API, or block-Thomas recurrence changes. Acceptance criterion: materially reduce coupled allocations/memory without changing results or regressing timing. Measurement/certification only; do not merge directly.


oameye:
Negative experiment: do not retain wrapper reuse.

Julia 1.10/LTS fails the direct dense-reference block-Thomas regression when `factors[i] = C` is removed. The computed RHS solution is numerically wrong, while the parent #418 recurrence passes this exact regression on both current Julia and LTS.

So the cached `Cholesky` wrapper cannot be assumed to remain a valid reusable factorization handle after refactorizing the underlying Schur matrix in place across supported Julia versions. This is a correctness failure, not a performance tradeoff.

Closing without merge. The production block-Thomas candidate will retain `factors[i] = C` and use the fully certified #418 unscaled recurrence.

